PR #72, opened 2023-11-08
add CI

iteyelmp:
Add CI to the project and verify that the code is correct when pushing the code.


.github/workflows/github-actions-es.yml
```diff
@@ -0,0 +1,23 @@
+# This workflow will build a golang project
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
+name: GitHub Actions ES
+run-name: ${{ github.actor }} is push code to main 🚀
+on:
+  push:
+    branches: [ "main", "ci" ]
```
qzhodl, line 7: does it mean that it only works on the ci branch merge into main branch?

ethstorage/p2p/protocol/sync_test.go
```diff
@@ -1,6 +1,8 @@
 // Copyright 2022-2023, EthStorage.
 // For license information, see https://github.com/ethstorage/es-node/blob/main/LICENSE
 
+//go:build !ci
```
qzhodl, line 4: Since you've add the !ci flag, I suggest that remove the related following changes in this file and leave it with Ping to  fix it

ethstorage/storage_manager_test.go
```diff
@@ -12,99 +14,98 @@ import (
 	"github.com/ethstorage/go-ethstorage/ethstorage/eth"
 )
 
-
```
qzhodl, line 15: The same as the sync test file

.github/workflows/github-actions-es.yml
iteyelmp: It will only be triggered when code is submitted to the main or ci branch.

ethstorage/storage_manager_test.go
iteyelmp: This is the automatic formatting of the code

qzhodl (approved):

Issue #74, opened 2023-11-09
We may need to avoid to let user manually assign a p2p advertise IP to generate a ENR that contains the public ip address


qzhodl:
Now if the users do not assign their public IP address in the command line flag and start es-node, it will not provide p2p sync service to other peers because the generate ENR doesn't contain the IP address.

PR #77, opened 2023-11-11
fix Issue74: use NAT ips as fallback ips if AdvertiseIP no be set

ping-ke:
related issue: https://github.com/ethstorage/es-node/issues/74

If the users do not assign their public IP address in the command line flag and start es-node, it will not provide p2p sync service to other peers because the generated ENR doesn't contain the IP address. 
As the node's host enables NAT service, we can use the host's addresses in the ENR if the AdvertiseIP is not set. 

How to verify:
Run the node without AdvertiseIP, and view the ENR at https://enr-viewer.com/. and see the IP in tcp4Multiaddr.


qzhodl:
Can we handle the Solo case which the host doesn't have a public ip address?

ethstorage/p2p/discovery.go
```diff
@@ -50,6 +51,19 @@ func (conf *Config) Discovery(log log.Logger, l1ChainID uint64, tcpPort uint16)
 	localNode := enode.NewLocalNode(conf.DiscoveryDB, priv)
 	if conf.AdvertiseIP != nil {
 		localNode.SetStaticIP(conf.AdvertiseIP)
+	} else {
+		for _, addr := range addrs {
+			ipStr, err := addr.ValueForProtocol(4)
```
qzhodl, line 56: May need to add a comment to use autoNAT for more advancing check

qzhodl (approved):

PR #78, opened 2023-11-13
Fix integration tests

syntrust:
Move environment-dependent tests to a folder `integration-tests` so they can run together locally but not in ci. 
Add integration_tests/run_tests.sh to get env and zkey ready and start tests.


qzhodl (approved):
LGTM

Issue #86, opened 2023-11-21
Filling empty in an external SSD on a MBP is super slow (160 hours)


qzhodl:
MBP configuration: 2.6 GHz 6-Core Intel Core i7

fio writing test result: 
`WRITE: bw=669MiB/s (701MB/s), 52.2MiB/s-56.3MiB/s (54.7MB/s-59.0MB/s), io=156GiB (167GB), run=238649-238650msec`

fio writing cmd: 
`fio --name=write-test --filename=/Volumes/PS2000/test.txt --ioengine=psync --direct=1 --bs=4k --rw=write --numjobs=12 --iodepth=1 --size=100g`

fio reading test result:
`READ: bw=145MiB/s (152MB/s), 12.0MiB/s-12.1MiB/s (12.6MB/s-12.7MB/s), io=7627MiB (7997MB), run=52684-52685msec`

fio reading cmd
`fio --name=read-test --filename=/Volumes/PS2000/test.txt --ioengine=psync --direct=1 --bs=4k --rw=randread --numjobs=12 --iodepth=1 --size=10g`

qzhodl:
After removing the "getLastKVIndex()" and "GetKvMetas" RPC calls, the filling empty process ran faster and could be finished in around 28 hours, and at the same time the CPU usage was 100%. 

Also, it means that the CPU on the tested MBP can only provide 10M/s encoding+KZG bandwidth.

In the next step we need to improve the RPC call latency in the syncing process.

qzhodl:
Fixed by https://github.com/ethstorage/es-node/pull/93, the direct io related issue is tracked by another issue.

Issue #85, opened 2023-11-20
investigate performance loss on 7950X and 5975WX


qizhou:
The filing performance on 7950X and TR 5975WX is much worse than ax101 (5950X).  Below are some numbers:

ax101:
<img width="1147" alt="image" src="https://github.com/ethstorage/es-node/assets/2541286/6a287930-ea40-4ba8-a5b5-ef8c4a442bd3">

~ 212 blobs/s

7950x:
<img width="1151" alt="image" src="https://github.com/ethstorage/es-node/assets/2541286/5f1420ee-827d-4306-9bdd-b90f78d9fd9b">

~ 36.0 blobs/s

<img width="1147" alt="image" src="https://github.com/ethstorage/es-node/assets/2541286/5841d64b-914f-420a-8ced-3611fbbc0d14">

~ 46 blobs/s

qizhou:
Need to check whether it is CPU issue or networking

qzhodl:
Should be fixed by https://github.com/ethstorage/es-node/pull/93

PR #93, opened 2023-11-23
Cache blob metas and lastKvIndex before blob syncing to improve the performance of p2p syncing

qzhodl:
Every time es-node sends a BLOB p2p sync request or a local empty blob-filling, it will send a getKvMetas/getLastKvIdx RPC call. When the RPC latency is high, it will increase a lot of syncing time, and people can check https://github.com/ethstorage/es-node/issues/85 and https://github.com/ethstorage/es-node/issues/86 for more details.

Due to the slow update rate of on-chain blob hashes and lastKvIdx (every two Ethereum mainnet epochs), we can optimize the process by initially downloading them into the local cache before syncing. Subsequently, they can be updated by the downloader every two epochs. This approach allows the peer-to-peer (p2p) sync to directly utilize the local cache, rather than repeatedly making RPC requests, thereby enhancing overall syncing performance.

The implementation includes the following parts:
1. The p2p sync client checks if the syncing is done, and will start to download the metas if syncing is not done.
2. The download process is multi-threading (32 threads), and it takes around 5 mins for 8,000,000 blobs in our office.
3. The downloader will update the local meta cache and lastKvIdx every 2 epochs
4. The p2p sync client only uses the local meta/lastKvIndex cache when it commits blobs into the storage file.

### DIFF
--- a/ethstorage/p2p/protocol/syncclient.go
+++ b/ethstorage/p2p/protocol/syncclient.go
@@ -129,9 +129,11 @@ type StorageManager interface {
 
 	StorageManagerWriter
 
-	LastKvIndex() (uint64, error)
+	LastKvIndex() uint64
 
 	DecodeKV(kvIdx uint64, b []byte, hash common.Hash, providerAddr common.Address, encodeType uint64) ([]byte, bool, error)
+
+	DownloadAllMetas() error
 }
 
 type SyncClient struct {
@@ -277,12 +279,7 @@ func (s *SyncClient) loadSyncStatus() {
 	}
 
 	// create tasks
-	lastKvIndex, err := s.storageManager.LastKvIndex()
-	if err != nil {
-		// TODO: panic?
-		log.Info("LoadSyncStatus failed: get lastKvIdx")
-		lastKvIndex = 0
-	}
+	lastKvIndex := s.storageManager.LastKvIndex()
 	for _, sid := range s.storageManager.Shards() {
 		exist := false
 		for _, task := range progress.Tasks {
@@ -448,7 +445,7 @@ func (s *SyncClient) cleanTasks() {
 	}
 }
 
-func (s *SyncClient) Start() {
+func (s *SyncClient) Start() error {
 	if s.startTime == (time.Time{}) {
 		s.startTime = time.Now()
 		s.logTime = time.Now()
@@ -456,9 +453,18 @@ func (s *SyncClient) Start() {
 
 	// Retrieve the previous sync status from LevelDB and abort if already synced
 	s.loadSyncStatus()
+	s.cleanTasks()
+	if !s.syncDone {
+		err := s.storageManager.DownloadAllMetas()
+		if err != nil {
+			return err
+		}
+	}
 
 	s.wg.Add(1)
 	go s.mainLoop()
+
+	return nil
 }
 
 func (s *SyncClient) AddPeer(id peer.ID, shards map[common.Address][]uint64) bool {
@@ -999,14 +1005,12 @@ func (s *SyncClient) FillFileWithEmptyBlob(start, limit uint64) (uint64, error) 
 		next     = start
 	)
 	defer s.metrics.ClientFillEmptyBlobsEvent(inserted, time.Since(st))
-	lastBlobIdx, err := s.storageManager.LastKvIndex()
-	if err != nil {
-		return start, fmt.Errorf("get lastBlobIdx for FillEmptyKV fail, err: %s", err.Error())
-	}
+	lastBlobIdx := s.storageManager.LastKvIndex()
+
 	if start < lastBlobIdx {
 		start = lastBlobIdx
 	}
-	inserted, next, err = s.storageManager.CommitEmptyBlobs(start, limit)
+	inserted, next, err := s.storageManager.CommitEmptyBlobs(start, limit)
 
 	return next, err
 }
--- a/ethstorage/p2p/protocol/syncclient.go
+++ b/ethstorage/p2p/protocol/syncclient.go
@@ -453,13 +453,6 @@ func (s *SyncClient) Start() error {
 
 	// Retrieve the previous sync status from LevelDB and abort if already synced
 	s.loadSyncStatus()
-	s.cleanTasks()
-	if !s.syncDone {
-		err := s.storageManager.DownloadAllMetas()
-		if err != nil {
-			return err
-		}
-	}
 
 	s.wg.Add(1)
 	go s.mainLoop()
@@ -567,6 +560,15 @@ func (s *SyncClient) RequestL2List(indexes []uint64) (uint64, error) {
 
 func (s *SyncClient) mainLoop() {
 	defer s.wg.Done()
+
+	s.cleanTasks()
+	if !s.syncDone {
+		err := s.storageManager.DownloadAllMetas()
+		if err != nil {
+			log.Error("Download blob metadata failed", "error", err)
+			return
+		}
+	}
 
 	for {
 		// Remove all completed tasks and terminate sync if everything's done
--- a/ethstorage/p2p/protocol/syncclient.go
+++ b/ethstorage/p2p/protocol/syncclient.go
@@ -133,7 +133,7 @@ type StorageManager interface {
 
 	DecodeKV(kvIdx uint64, b []byte, hash common.Hash, providerAddr common.Address, encodeType uint64) ([]byte, bool, error)
 
-	DownloadAllMetas() error
+	DownloadAllMetas(batchSize uint64) error
 }
 
 type SyncClient struct {
@@ -563,7 +563,7 @@ func (s *SyncClient) mainLoop() {
 
 	s.cleanTasks()
 	if !s.syncDone {
-		err := s.storageManager.DownloadAllMetas()
+		err := s.storageManager.DownloadAllMetas(s.syncerParams.MetaDownloadBatchSize)
 		if err != nil {
 			log.Error("Download blob metadata failed", "error", err)
 			return
--- a/ethstorage/storage_manager.go
+++ b/ethstorage/storage_manager.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"math/big"
 	"sync"
 	"time"
@@ -409,6 +410,7 @@ func (s *StorageManager) downloadMetaInParallel(from, to uint64) error {
 }
 
 func (s *StorageManager) downloadMetaInRange(from, to uint64, taskId uint64) error {
+	rangeStart := from
 	for from < to {
 		batchLimit := from + MetaBatchSize
 		if batchLimit > to {
@@ -442,7 +444,12 @@ func (s *StorageManager) downloadMetaInRange(from, to uint64, taskId uint64) err
 		}
 		s.mu.Unlock()
 
-		log.Info("One batch metas has been downloaded", "first", from, "batchLimit", batchLimit, "to", to, "taskId", taskId)
+		log.Info(
+			"One batch metas has been downloaded", "first", from,
+			"batchLimit", batchLimit,
+			"to", to,
+			"progress", fmt.Sprintf("%.1f%%", float64((from - rangeStart)*100)/float64(to - rangeStart)),
+			"taskId", taskId)
 
 		from = batchLimit
 	}
--- a/ethstorage/storage_manager.go
+++ b/ethstorage/storage_manager.go
@@ -5,12 +5,16 @@ package ethstorage
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"math/big"
 	"sync"
+	"time"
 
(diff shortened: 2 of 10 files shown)


ethstorage/storage_manager.go
```diff
@@ -293,6 +302,143 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
+
+		log.Info("One batch metas has been downloaded", "first", from, "batchLimit", batchLimit, "to", to, "taskId", taskId)
+
+		from = batchLimit
+	}
+	return nil
+}
+
+func (s *StorageManager) updateLocalMetas(kvIndices []uint64, commits []common.Hash) {
+	for i, idx := range kvIndices {
```
qizhou, line 425: Lock is required?

ethstorage/storage_manager.go
```diff
@@ -293,6 +302,143 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
+		meta := [32]byte{}
+		new(big.Int).SetInt64(int64(idx)).FillBytes(meta[0:5])
+		copy(meta[32-HashSizeInContract:32], commits[i][0:HashSizeInContract])
+
+		s.blobMetas[idx] = meta
+	}
+}
+
+func (s *StorageManager) getKvMetas(kvIndices []uint64) [][32]byte {
+	metas := [][32]byte{}
```
qizhou, line 435: We need to acquire s.metaMapMu.Lock() for reading (basically any access).

ethstorage/storage_manager.go
```diff
@@ -26,22 +32,27 @@ type Il1Source interface {
...
 // and a consistent view of most-recent-finalized L1 block.
 type StorageManager struct {
 	shardManager      *ShardManager
-	mu                sync.Mutex // protect localL1 and underlying blob read/write
-	localL1           int64      // local view of most-recent-finalized L1 block
+	mu                sync.Mutex // protect lastKvIdx, shardManager and blobMeta read/write state
+	lastKvIdx         uint64     // lastKvIndex in the most-recent-finalized L1 block
 	l1Source          Il1Source
 	DownloadThreadNum int
+	metaMapMu         sync.Mutex // this is only for blobMetas writing thread safe in downloadMetaInParallel
```
qizhou, line 47: Both read/write need the lock, and we should comment that we allow lock sequence like mu -> metaMapMu, but never metaMapMu -> mu to avoid dead lock.

ethstorage/storage_manager.go
```diff
@@ -293,6 +302,143 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
 	return nil
 }
 
+// DownloadAllMetas This function download the blob hashes of all the local storage shards from the smart contract
+func (s *StorageManager) DownloadAllMetas() error {
+	s.mu.Lock()
```
qizhou, line 307: This will lock other threads that accessing mu.Lock() - since the call of this method can be costly, it is intentional?

ethstorage/storage_manager.go
```diff
@@ -293,6 +302,143 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
+		for i := from; i < batchLimit; i++ {
+			kvIndices = append(kvIndices, i)
+		}
+
+		metas, err := s.l1Source.GetKvMetas(kvIndices, l1)
+		if err != nil {
+			return err
+		}
+
+		s.metaMapMu.Lock()
```
qizhou, line 411: I guess you can reuse mu without introducing metaMapMu.  Further, it should check whether l1 has changed when writing it into the map to avoid race conditions.

qizhou:
I can successfully run this on my MacBook, which has a poor connection with L1.  Great job!

A couple of minor improvements:
- It will be great to report a progress (e.g., %) during meta downloading
- GetKvMetas may fail occasionally and a retry will be helpful before returning an error!

qzhodl:
> I can successfully run this on my MacBook, which has a poor connection with L1. Great job!
> 
> A couple of minor improvements:
> 
> * It will be great to report a progress (e.g., %) during meta downloading
> * GetKvMetas may fail occasionally and a retry will be helpful before returning an error!

Fixed

ethstorage/storage_manager.go
```diff
@@ -5,17 +5,24 @@ package ethstorage
...
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rpc"
 )
 
 const (
 	blobFillingMask    = byte(0b10000000)
 	HashSizeInContract = 24
+	MetaBatchSize      = 8000
```
qizhou, line 24: Any comments on using the value of the batch size?

ethstorage/storage_manager.go
```diff
@@ -104,12 +116,23 @@ func (s *StorageManager) DownloadFinished(newL1 int64, kvIndices []uint64, blobs
...
+	if err != nil {
+		s.mu.Unlock()
+		return err
+	}
+	s.lastKvIdx = lastKvIdx
 	s.localL1 = newL1
 
+	s.mu.Unlock()
+
+	s.updateLocalMetas(kvIndices, commits)
```
qizhou, line 134: It is generally not a good practice to drop a lock during an atomic operation.

ethstorage/storage_manager.go
```diff
@@ -293,6 +321,164 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
+			kvIndices = append(kvIndices, i)
+		}
+
+		s.mu.Lock()
+		localL1 := s.localL1
+		s.mu.Unlock()
+
+		metas, err := s.l1Source.GetKvMetas(kvIndices, localL1)
+		if err != nil {
+			// Retry the request again in case it could fail occasionally in poor network connection
```
qizhou, line 429: A repeated retry with a delay (e.g., sleep 2s for 10 times) will be better here.

ethstorage/storage_manager.go
```diff
@@ -293,6 +321,164 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
+	}
+}
+
+func (s *StorageManager) getKvMetas(kvIndices []uint64) [][32]byte {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	metas := [][32]byte{}
+	for _, i := range kvIndices {
+		metas = append(metas, s.blobMetas[i])
```
qizhou, line 477: It is possible that `i` does not exist in `blobMetas`?

ethstorage/p2p/protocol/syncclient.go
```diff
@@ -448,17 +445,26 @@ func (s *SyncClient) cleanTasks() {
...
 	if s.startTime == (time.Time{}) {
 		s.startTime = time.Now()
 		s.logTime = time.Now()
 	}
 
 	// Retrieve the previous sync status from LevelDB and abort if already synced
 	s.loadSyncStatus()
+	s.cleanTasks()
+	if !s.syncDone {
+		err := s.storageManager.DownloadAllMetas()
```
qizhou, line 458: I think you can call DownloadAllMetas() in goroutine as we shouldn't block `Start()`

ethstorage/storage_manager.go
```diff
@@ -293,6 +321,164 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
 
+// DownloadAllMetas This function download the blob hashes of all the local storage shards from the smart contract
+func (s *StorageManager) DownloadAllMetas() error {
+	header, err := s.l1Source.HeaderByNumber(context.Background(), big.NewInt(rpc.FinalizedBlockNumber.Int64()))
+	if err != nil {
+		return err
+	}
+	l1 := header.Number.Int64()
+
+	lastKvIdx, err := s.l1Source.GetStorageLastBlobIdx(l1)
```
qizhou, line 332: I believe it is safer to retrieve lastKvIdx and l1 from local variables with lock.

ethstorage/storage_manager.go
```diff
@@ -293,6 +320,163 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
+		kvIndices := []uint64{}
+		for i := from; i < batchLimit; i++ {
+			kvIndices = append(kvIndices, i)
+		}
+
+		s.mu.Lock()
+		localL1 := s.localL1
+		s.mu.Unlock()
+
+		metas, err := s.l1Source.GetKvMetas(kvIndices, localL1)
```
qizhou, line 419: This method may fail forever if remove is supported and `lastKvIndex` is decreased.  A better way would be
- (with lock) get localL1 and `lastKvIndex`
- reduce `to` and `batchLimit` if they are larger than `lastKvIndx`
- call GetKvMetas

ethstorage/storage_manager.go
```diff
@@ -108,8 +114,15 @@ func (s *StorageManager) DownloadFinished(newL1 int64, kvIndices []uint64, blobs
...
 	}
 
+	lastKvIdx, err := s.l1Source.GetStorageLastBlobIdx(newL1)
+	if err != nil {
+		return err
+	}
+	s.lastKvIdx = lastKvIdx
 	s.localL1 = newL1
 
+	s.updateLocalMetas(kvIndices, commits)
```
qizhou, line 124: We should remove the metas in the map if lastKvIdx is smaller because of removal (or at least a TODO here).

ethstorage/storage_manager.go
```diff
@@ -154,20 +174,22 @@ func (s *StorageManager) CommitBlobs(kvIndices []uint64, blobs [][]byte, commits
...
 
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	metas, err := s.l1Source.GetKvMetas(kvIndices, s.localL1)
+	metas, err := s.getKvMetas(kvIndices)
 	if err != nil {
 		return nil, err
 	}
+
+	s.mu.Lock()
```
qizhou, line 182: The KvMetas can be changed after calling line 177 with lock dropped.  We should stick to the previous code logic from line 157 and acquire the lock from `getKvMetas` and to the rest.

ethstorage/storage_manager.go
```diff
@@ -196,18 +218,20 @@ func (s *StorageManager) CommitEmptyBlobs(start, limit uint64) (uint64, uint64,
 		kvIndices = append(kvIndices, i)
 	}
 
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	metas, err := s.l1Source.GetKvMetas(kvIndices, s.localL1)
+	metas, err := s.getKvMetas(kvIndices)
```
qizhou, line 221: Similar the aforementioned race conditions can happen.

qizhou (commented):
A lot of good changes.  Just a few comments below.

ethstorage/storage_manager.go
```diff
@@ -293,6 +322,175 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
+}
+
+// Please note that the caller function must uses s.mu to protect the s.blobMetas reading in this function
+func (s *StorageManager) getKvMetas(kvIndices []uint64) ([][32]byte, error) {
+	metas := [][32]byte{}
+	for _, i := range kvIndices {
+		meta, ok := s.blobMetas[i]
+		if ok {
+			metas = append(metas, meta)
+		} else {
```
qizhou, line 487: I think we should do more sanity checks here:
- If i >= lastKvIdx, then return empty metas
- else return error (which is not expected to happen).

ethstorage/storage_manager.go
```diff
@@ -293,6 +322,175 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
+		meta := [32]byte{}
+		new(big.Int).SetInt64(int64(idx)).FillBytes(meta[0:5])
+		copy(meta[32-HashSizeInContract:32], commits[i][0:HashSizeInContract])
+
+		s.blobMetas[idx] = meta
+	}
+
+	// In case the lastKvIdx is smaller than oldLastKvIdx because of removal, we need to reset the removal metas
+	for i := s.lastKvIdx; i < oldLastKvIndex; i++ {
+		meta := [32]byte{}
```
qizhou, line 473: I think we should delete these metas for consistency.

ethstorage/storage_manager.go
```diff
@@ -293,6 +320,168 @@ func (s *StorageManager) syncCheck(kvIdx uint64) error {
...
+}
+
+// Please note that the caller function must uses s.mu to protect the s.blobMetas reading in this function
+func (s *StorageManager) getKvMetas(kvIndices []uint64) ([][32]byte, error) {
+	metas := [][32]byte{}
+	for _, i := range kvIndices {
+		meta, ok := s.blobMetas[i]
+		if ok {
+			metas = append(metas, meta)
+		} else {
```
qizhou, line 472: nit: I think you can use  if {} else if {} else { return nil, error}

qizhou (approved):

Issue #84, opened 2023-11-20
add pre-build versions of es-node


qizhou:
Similar to geth, we should provide pre-build versions so that the node operator is not required to go through the detailed steps to run a node.

syntrust:
Solved in https://github.com/ethstorage/es-node/pull/134

PR #94, opened 2023-11-23
Add github workflow to release with pre-build versions of es-node

syntrust:
address https://github.com/ethstorage/es-node/issues/84

Leave the doc update to https://github.com/ethstorage/es-node/pull/92

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Build
         run: |
           make
-          mv es-node es-node.${{github.ref_name}}.ubuntu
+          mv cmd/es-node/es-node es-node.${{github.ref_name}}.ubuntu
 
       - name: Create Release
         uses: softprops/action-gh-release@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,32 @@
...
+
+name: GitHub Actions ES Publish
+run-name: ${{ github.actor }} is publishing a release 🚀
+on:
+  push:
+    tags: 
+      - 'v*'
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
qzhodl, line 11: If github also support darwin and windows, we may also want to support it. Also is it x86 64-bit build? Some people   may run es-node in their MBP with ARM64, so it will be nice to support it too.

syntrust: now we support:

- es-node.v0.1.2.darwin-amd64 
- es-node.v0.1.2.darwin-arm64
- es-node.v0.1.2.linux-amd64 
- es-node.v0.1.2.windows-amd64

qzhodl (approved):

PR #111, opened 2023-12-01
add RecordBandwidth metrics

ping-ke:
add RecordBandwidth metrics


ethstorage/p2p/node.go
```diff
@@ -87,10 +82,6 @@ func (n *NodeP2P) init(resourcesCtx context.Context, rollupCfg *rollup.EsConfig,
 			n.gater = extra.ConnectionGater()
 			n.connMgr = extra.ConnectionManager()
 		}
-		m := (protocol.Metricer)(nil)
-		if rollupCfg.MetricsEnable {
```
qzhodl, line 91: Is there anyway we can enable metric after this change? Plus, I would say this PR has lower priority since we have several high issues pending from my point of view

qzhodl (approved):

PR #115, opened 2023-12-01
Build docker image, etc.

syntrust:
1. release zip with scripts and pre-built binaries
2. build a Docker image and push it to  GitHub Container Registry (ghcr.io) on each release
    - can be managed in https://github.com/orgs/ethstorage/packages
3. move the built binary (es-node) and snark folders(snarkjs, snarkbuild) to top level so that does not depend on src structures.

Note that the corresponding documentation will be changed in https://github.com/ethstorage/es-node/pull/92


ethstorage/miner/config.go
```diff
@@ -38,6 +37,6 @@ var DefaultConfig = Config{
 	GasPrice:         nil,
 	PriorityGasPrice: nil,
 	ZKeyFileName:     "blob_poseidon.zkey",
-	ZKWorkingDir:     filepath.Join("ethstorage", "prover"),
+	ZKWorkingDir:     "./",
```
qzhodl, line 40: It will break the case that users compile and run es-node directly

syntrust:
closed because the branch has some big files submitted accidentally

Issue #112, opened 2023-12-01
es-node init command should support user-specific file path, blob start index and length


qzhodl:
To support multiple storage data files for one shard.

PR #130, opened 2023-12-04
(Draft) Init data files with blob range per file

syntrust:


syntrust:
Closed as the priority of the requirements has been lowered: https://github.com/ethstorage/es-node/issues/112

Issue #131, opened 2023-12-05
Cannot add external IP to the ENR if the es-node behind a NAT and advertise IP is not set


ping-ke:
Cannot add external IP to the ENR if the es-node behind a NAT and advertise IP is not set, so other peers cannot directly connect it to that es-node using that ENR.

Related issue: https://github.com/ethstorage/es-node/issues/74

PR #132, opened 2023-12-05
Set external IP and mapping port to es-node  ENR when it behind a NAT and the advertise IP is not set

ping-ke:
Issue: https://github.com/ethstorage/es-node/issues/131
If the es-node is running behind a NAT and without the advertiseIP being set. we can discover the NAT service and add port mappings to the NAT service, then add the external IP and mapping ports to ENR.

How to test:
Run the es-node behind NAT service without advertiseIP being set. Then copy the ENR the es-node output and check it in the https://enr-viewer.com/. 
`enr:-Li4QCVR_onX8sRp5LVSxHTMbxIkvJVcS3VXu9c3DcFBhnrjK4YmZHc55Mhktqi9Qj-ypbbUms_u56PyrXbBT2po7CCGAYwvPP-9imV0aHN0b3JhZ2XbAYDY15Sfn1_YmtZI8sAAyVTY2ch3QyQ-xcGAgmlkgnY0gmlwhEmqKEKJc2VjcDI1NmsxoQP9Mik329t5ntsy0m8DWu-JRgT6T1TNOCx_M62HV9GPCIN0Y3CCggGDdWRwgnZc`

we can see its multiaddr like
```
tcp4 multiaddr  /ip4/73.170.40.66/tcp/33281/p2p/16Uiu2HAmVhMrt2PoKbxBjEoFPzc3auqn4HM82oqZrXwkpUAeGgEK
udp4 multiaddr  /ip4/73.170.40.66/udp/30300/p2p/16Uiu2HAmVhMrt2PoKbxBjEoFPzc3auqn4HM82oqZrXwkpUAeGgEK
```

run the second es-node with the first es-node ENR and see the log like the following
```
t=2023-12-04T23:54:39+0800 lvl=info msg="Connected to peer"                     peer=16Uiu2HAmVhMrt2PoKbxBjEoFPzc3auqn4HM82oqZrXwkpUAeGgEK addr=/ip4/73.170.40.66/tcp/32501
... ...
t=2023-12-04T23:55:06+0800 lvl=info msg="Disconnected from peer"                peer=16Uiu2HAmVhMrt2PoKbxBjEoFPzc3auqn4HM82oqZrXwkpUAeGgEK addr=/ip4/73.170.40.66/tcp/32501
```


ethstorage/p2p/discovery.go
```diff
@@ -149,6 +166,30 @@ func (v *Secp256k1) DecodeRLP(s *rlp.Stream) error {
 	return nil
 }
 
+func addNATMappings(tcpPort, udpPort int) (net.IP, int, int, error) {
```
qzhodl, line 169: We may want to investigate whether libp2p can do this work if we enable NAT option in the libp2p constructor

ethstorage/p2p/discovery.go
```diff
@@ -62,23 +78,24 @@ func (conf *Config) Discovery(log log.Logger, l1ChainID uint64, tcpPort uint16,
...
-		localNode.Set(enr.TCP(conf.ListenTCPPort))
-	} else {
-		return nil, nil, fmt.Errorf("no TCP port to put in discovery record")
+
+		if !end {
+			intTcpPort := localNode.Node().TCP()
+			intUdpPort := localNode.Node().UDP()
+			ip, extTcpPort, extUdpPort, err := addNATMappings(intTcpPort, intUdpPort)
+			if err == nil {
+				localNode.SetStaticIP(ip)
```
qzhodl, line 90: what if the IP is still an internal IP address?

ethstorage/p2p/discovery.go
ping-ke: NAT manager is not public, we cannot directly use it.

PR #148, opened 2023-12-13
Estimate mining profit

syntrust:
Estimate tx cost and mining reward and compare the profit to the min-profit flag before submitting mining txes.
- implement miner's reward computation based on the logic of the smart contract
- read the actual cost and reward from tx receipt for users to check.
- remove gas price related settings from run.sh so that a market price can be used.

### DIFF
--- a/ethstorage/miner/l1_mining_api.go
+++ b/ethstorage/miner/l1_mining_api.go
@@ -19,7 +19,10 @@ import (
 	"github.com/ethstorage/go-ethstorage/ethstorage/eth"
 )
 
-const gasBufferRatio = 1.2
+const (
+	gasBufferRatio    = 1.2
+	rewardDenominator = 10000
+)
 
 func NewL1MiningAPI(l1 *eth.PollingClient, lg log.Logger) *l1MiningAPI {
 	return &l1MiningAPI{l1, lg}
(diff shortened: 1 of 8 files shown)


ethstorage/miner/worker.go
```diff
@@ -344,11 +349,32 @@ func (w *worker) checkTxStatus(txHash common.Hash, miner common.Address) {
...
+		log.Info("Mining transaction details", "txHash", txHash, "gasUsed", receipt.GasUsed, "effectiveGasPrice", receipt.EffectiveGasPrice)
+		cost := new(big.Int).Mul(new(big.Int).SetUint64(receipt.GasUsed), receipt.EffectiveGasPrice)
+		var reward *big.Int
+		for _, rLog := range receipt.Logs {
+			if rLog.Topics[0] == minedEventSig {
+				// the last param of total unindexed 3
+				reward = new(big.Int).SetBytes(rLog.Data[64:])
+			}
+		}
+		log.Info("Mining transaction accounting (in ether)", "reward", weiToEther(reward), "cost", weiToEther(cost), "profit", weiToEther(new(big.Int).Sub(reward, cost)))
```
qzhodl, line 361: It crashed on one of my node when running this PR

qzhodl (approved):
LGTM

ethstorage/miner/l1_mining_api.go
```diff
@@ -142,6 +135,34 @@ func (m *l1MiningAPI) SubmitMinedResult(ctx context.Context, contract common.Add
 		return common.Hash{}, fmt.Errorf("failed to estimate gas: %w", err)
 	}
 	m.lg.Info("Estimated gas done", "gas", estimatedGas)
+	cost := new(big.Int).Mul(new(big.Int).SetUint64(estimatedGas), gasPrice)
```
ping-ke, line 138: I think we can change the strategy: if mining info changes, start estimating which block we can start to have profit, then start mining from that block.

qzhodl: Maybe for every new block, we can first estimate the profit before beginning to mine, and if it doesn't meet the requirement, then we can skip this block? But it also depends on the dynamic gas price when we are estimating.

syntrust: Tracked by https://github.com/ethstorage/es-node/issues/151

qzhodl (approved):

ethstorage/miner/l1_mining_api.go
ping-ke: resolve as will be updated later (https://github.com/ethstorage/es-node/issues/151)

ping-ke (approved):

Issue #168, opened 2023-12-27
Should support `--p2p.disable`


syntrust:
# Rationale

Why should this feature exist?
1. We have flag `--p2p.disable` documented. 
2. Too much log when peer not exist

> INFO [12-27|15:55:15.911] Close connection as AddPeer fail         peer=16Uiu2HAm4n64b2uQnYYD2wrFkQQVi1f1fpNozpP2Aqqws6MQhYG6
> INFO [12-27|15:55:15.914] Connected to peer                        peer=16Uiu2HAm4n64b2uQnYYD2wrFkQQVi1f1fpNozpP2Aqqws6MQhYG6 addr=/ip4/62.72.46.81/tcp/9222
> INFO [12-27|15:55:15.914] Cannot remove peer from sync duties, peer was not registered peer=16Uiu2HAm4n64b2uQnYYD2wrFkQQVi1f1fpNozpP2Aqqws6MQhYG6
> INFO [12-27|15:55:15.914] Disconnected from peer                   peer=16Uiu2HAm4n64b2uQnYYD2wrFkQQVi1f1fpNozpP2Aqqws6MQhYG6 addr=/ip4/62.72.46.81/tcp/9222
> INFO [12-27|15:55:16.174] No need this peer, the connection would be closed later peer=16Uiu2HAkym6S9pc26eLeoKkao1z4EAoKXY8ysV7n9Am4kXoWAjse shards=map[0xb4B46bdAA835F8E4b4d8e208B6559cD267851051:[0]]

What are the use-cases?

Test mining, etc.

# Implementation
Do you have ideas regarding the implementation of this feature?

Use flag `--p2p.disable` in run.sh

Are you willing to implement this feature?

We have the flag in doc, but use it causes panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x100ee812f]

goroutine 1 [running]:
github.com/ethstorage/go-ethstorage/ethstorage/p2p.(*NodeP2P).Start(...)
        /Users/dl/code/es-node/ethstorage/p2p/node.go:241
github.com/ethstorage/go-ethstorage/ethstorage/node.(*EsNode).Start(0xc0008e6480, {0x1014b4f2b?, 0x1?}, 0xc000302900?)
        /Users/dl/code/es-node/ethstorage/node/node.go:255 +0x10f
main.EsNodeMain(0xc0005d14a0?)
[...]

PR #170, opened 2023-12-28
Issue168: fix es-node crash when --p2p.disable set

ping-ke:
related issue: https://github.com/ethstorage/es-node/issues/168

How to test:
1. Run p2p sync_test.go
2. Run es-node with `--p2p.disable` param, and es-node can start and close successfully.


syntrust (commented):
I wonder how this change will affect miner to actually start mining? Will sync done event still emitted?

ethstorage/storage_manager.go
```diff
@@ -153,6 +153,10 @@ func (s *StorageManager) Reset(newL1 int64) error {
 	return nil
 }
 
+func (s *StorageManager) GetLastKVIdxByBlockNumber(blockNumber int64) (uint64, error) {
```
qzhodl, line 156: what do we need to do this change？

qzhodl:
> I wonder how this change will affect miner to actually start mining? Will sync done event still emitted?

Maybe the miner module can have another flag to indicate start at launch without waiting for the signal?

ethstorage/storage_manager.go
syntrust: I think it's ok as long as the behavior is documented

syntrust (approved):

README.md
```diff
@@ -91,7 +91,7 @@ The full list of options that you can use to configure an es-node are as follows
 |`--p2p.advertise.udp`|The UDP port to advertise in Discv5 as fallback if not determined by Discv5, put into the ENR of the node. Set to p2p.listen.udp value if 0.|`0`||
 |`--p2p.ban.peers`|Enables peer banning. This should ONLY be enabled once certain peer scoring is working correctly.|||
 |`--p2p.bootnodes`|Comma-separated base64-format ENR list. Bootnodes to start discovering other node records from.|||
-|`--p2p.disable`|Completely disable the P2P stack|||
+|`--p2p.disable`|Completely disable the P2P stack, if p2p is disabled, mining will also be disabled.|||
```
qzhodl, line 94: I think the more accurate expression maybe "If P2P is disabled, mining will not start automatically"?

PR #190, opened 2024-01-15
fix download zkey

syntrust:
Looks like the old way to bypass the confirmation page does not work anymore..


run.sh
```diff
@@ -57,17 +57,15 @@ echo "zk prover mode is $zkp_mode"
...
+  zkey_url="https://drive.usercontent.google.com/download?id=1ZLfhYeCXMnbk6wUiBADRAn1mZ8MI_zg-&export=download&confirm=t&uuid=16ddcd58-2498-4d65-8931-934df3d0065c&at=APZUnTVkjg3SRn910o_UXqeYqOoy:1705311918635"
 fi
 zkey_file="./build/bin/snarkjs/$zkey_name"
 if [ ! -e  ${zkey_file} ]; then
-  echo "${zkey_file} not found, start downloading..."
-  html=`curl -c ./cookie -s -L "https://drive.google.com/uc?export=download&id=${file_id}"`
-  curl -Lb ./cookie "https://drive.google.com/uc?export=download&`echo ${html}|grep -Eo 'confirm=[a-zA-Z0-9\-_]+'`&id=${file_id}" -o ${zkey_file}
-  rm cookie
+  echo "${zkey_file} not found, start downloading..." 
+  curl $zkey_url -o ${zkey_file}
```
qzhodl, line 68: can we add a double check here to see if the zkey file is downloaded successfully, something like the file size > 100M

qzhodl (approved):

PR #194, opened 2024-01-18
doc update for devnet3

syntrust:


GUIDE.md
```diff
@@ -160,4 +160,17 @@ nvm use 20
 ### Install snarkjs
 ```sh
 npm install -g snarkjs
-```
\ No newline at end of file
+```
+
```
qzhodl, line 164: We may want to introduce a little what will the users are expected to go through after the es-node launch. For example, they will basically go through two phases:
1. Data sync phase: they can tell how many peers are connected, and more importantly, the estimated syncing time. we can only explain a little bit about why the CPUs are full, and how to tune it using p2p.sync.concurrency & p2p.fill-empty.concurrency. (Or move it to FAQ)
2. Sampling phase: they may need to find out whether the es-node can do enough sampling in one slot, and if the log shows " Mining tasks timed out", what should they do? Check the read IOPS of their disk, and they can also tune it by using --miner.threads-per-shard

GUIDE.md
```diff
@@ -160,4 +160,17 @@ nvm use 20
...
+
+## FAQ
+
```
qzhodl, line 167: Several more FAQ:
1. what should I do when the log frequently shows "i/o deadline reached" during data syncing?
2. what does it means when the log shows "The nonces are exhausted in this slot"
3. How do I know whether I've got a mining reward? (1. dashboard 2. related log 3. miner address balance)

GUIDE.md
```diff
@@ -160,4 +160,90 @@ nvm use 20
...
+### What should I do when the log frequently shows "i/o deadline reached" during data syncing?
+
+The timeout issue could be due to network reasons. It may be because the amount of data requested relative to network performance is too large, so even a slight delay in individual requests will be magnified.
+
+You can change the size of each request to a smaller value using `--p2p.max.request.size`. 
+The default value is `4194304`. You can try adjusting it to `1048576`.
+
+### How to tune the performance of syncing?
```
qzhodl, line 207: One related FAQ here: Why is the CPU fully utilized when syncing data? Does this imply that running an es-node requires high-end hardware? Additionally, is there a way to configure it to use less CPU power? I'm asking because I need to run other software on the same machine.

qzhodl (approved):

PR #199, opened 2024-01-22
Run.sh: new storage contract + specify udp port + new ENR

qzhodl:


run.sh
```diff
@@ -102,9 +102,10 @@ es_node_start=" --network devnet \
   --l1.beacon-based-time 1705546368 \
   --l1.beacon-based-slot 7419864 \
   --download.thread 32 \
+  --p2p.listen.udp 30305 \
```
syntrust, line 105: We may also need to change run-rpc.sh, run-docker.sh, docker-compose.yaml, and Dockerfile

Issue #164, opened 2023-12-21
A new node can not discover all the active peers that boot node connected


qzhodl:
On devnet-2, the bootnode was connected with 13 active peers, then we launched a new node, and it can only connect with 7 active peers. We may need to investigate the root cause.

qzhodl:
ip <> peer ids
![21291703847634_ pic](https://github.com/ethstorage/es-node/assets/5291653/d406aea4-0228-44ef-b86e-9db69eb2269e)

ping-ke:
The es-node uses UDP to broadcast and receive ENR and then puts the received ENRs (nodes) into the disc_v5 routing table. In `Discovery. go` of es-node, the `DiscoveryProcess ` puts the nodes in the routing table into Peerstore (if there is Peer information retained in the previous startup in esnode_PeerStore_db, it will also be loaded into PeerStore) and then the libp2p will connect nodes in the Peerstore through TCP. 
But if a node does not have a public IP and p2p.advertise.ip is not set, then there is no IP in its ENR, so when other nodes receive it and add it to disc v5, it cannot successfully put it into the routing table after receiving its ENR. See the code below.
` https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/table.go line#536`
```
func (tab *Table) addSeenNode(n *node) {
	... ...
	if !tab.addIP(b, n.IP()) {
		// Can't add: IP limit reached.
		return
	}
```

Therefore, for nodes with public IP, **ALL** nodes can broadcast their ENR to **ALL** nodes through routing table, so their Peer number is relatively high. Nodes without public IP can only obtain other nodes through the routing table (only contain nodes with public IP), but cannot connect to other nodes which only have private IP, so there is no connection between private IP nodes.
The Addresses currently in the routing table are:
```
5.9.87.214:30305 
88.99.30.186:30305 
65.109.62.253:30305 
65.109.115.36:30305 
213.133.99.85:30305 
65.109.50.145:30305 
62.72.46.81:30305 
```
Other nodes: 
```
73.170.40.66:9222    
65.108.234.173:9222  
65.108.236.27:9222   
142.132.208.253:9222 
119.86.43.247:19457  
162.219.87.212:9222 
```

qzhodl:
65.108.236.27 is one of our AX101 which has a public IP, but why it is in the "other nodes"?

ping-ke:
This because it is run in the docker which do not have public IP in the docker.

zhang-wenchao:
[...]

PR #200, opened 2024-01-23
Issue164: A new node can not discover all the active peers that boot node connected

ping-ke:
Related issues: https://github.com/ethstorage/es-node/issues/164, https://github.com/ethstorage/es-node/issues/131

if the node does not have AdvertiseIP and public IP (like run in the docker or behind a nat), we can get the public IP and the TCP port from the libp2p. host's addresses (get from the nat service or address observed by peers) and set to the node discovery. As the host's addresses would not show when it start, so we will check it every initLocalNodeAddrInterval. After the public IP and the TCP port are initialized, we will also refresh the address in case the address changes.

In this change, I fork go-ethereum to change the iptrackMinStatements (in go-ethereum\p2p\enode\localnode.go) from 10 to 4, which is used by IPTracker to predict the external endpoint, i.e. IP address and port, of the local host. (If the number of remote peers sees the same IP address and port, the IP and port will be set to localnode)

Test sample:
The following are our three nodes that run in docker, the machine behind NAT, and the machine running in Chinese. `All node` means a node in the routing table, and `peer` means peers in the sync client.
 
b79685f722c443f27860d9963ce56d5
d55c1c6189138a9cb30192c3d6420a6

How to test:
Run two es-nodes behind NAT service or docker without advertiseIP being set. and the two nodes can connect after some time.

How to verify in the real network:
If you are running a private node, you can compare its peer count with a public node's peer count, if the values are similar, it should work fine. if the peer count of the private node is much smaller than the public node, then we need more investigation.


ethstorage/p2p/discovery.go
```diff
@@ -348,9 +393,9 @@ func (n *NodeP2P) DiscoveryProcess(ctx context.Context, log log.Logger, l1ChainI
 			// or peers on different chains, or anyone we have not seen via discovery.
 			// There is no tag score decay yet, so just set it to 42.
 			n.ConnectionManager().TagPeer(info.ID, fmt.Sprintf("ethstorage-%d-%d", dat.ChainID, dat.Version), 42)
-			log.Debug("Discovered peer", "peer", info.ID, "nodeID", found.ID(), "addr", info.Addrs[0])
+			log.Debug("Discovered peer", "peer", info.ID, "nodeID", node.ID(), "addr", info.Addrs[0])
 		case <-connectTicker.C:
-			connected := n.Host().Network().Peers()
+			connected := n.syncCl.Peers()
```
qzhodl, line 398: Why do we do this change?

ethstorage/p2p/discovery.go
```diff
@@ -327,11 +371,12 @@ func (n *NodeP2P) DiscoveryProcess(ctx context.Context, log log.Logger, l1ChainI
 			log.Info("Stopped peer discovery")
 			return // no ctx error, expected close
 		case found := <-randomNodesCh:
+			node := n.dv5Udp.Resolve(found)
```
qzhodl, line 374: We may need some comment for this line

ethstorage/p2p/discovery.go
```diff
@@ -114,15 +113,38 @@ func (conf *Config) Discovery(log log.Logger, l1ChainID uint64, tcpPort uint16,
...
+}
+
+func updateLocalNodeIPAndTCP(addrs []ma.Multiaddr, localNode *enode.LocalNode) bool {
+	for _, addr := range addrs {
+		ipStr, err := addr.ValueForProtocol(4)
+		if err != nil {
+			continue
+		}
+		ip := net.ParseIP(ipStr)
+		if ip.IsPrivate() || !ip.IsGlobalUnicast() {
```
qzhodl, line 134: Maybe we can reuse getLocalPublicIPv4?

ping-ke: this function is to find out public IP from the node addresses (observed peers) for private node which is not equal to local public IP.

ethstorage/p2p/discovery.go
```diff
@@ -327,11 +371,13 @@ func (n *NodeP2P) DiscoveryProcess(ctx context.Context, log log.Logger, l1ChainI
 			log.Info("Stopped peer discovery")
 			return // no ctx error, expected close
 		case found := <-randomNodesCh:
+			// get the most recent version of the node record in case it change deal to the remote node TCP or UDP port change.
+			node := n.dv5Udp.Resolve(found)
```
qzhodl, line 375: As a node with a public IP, how can we verify that a node running in a docker can be accessed through the correct ENR?

ethstorage/p2p/discovery.go
```diff
@@ -258,16 +280,38 @@ func (n *NodeP2P) DiscoveryProcess(ctx context.Context, log log.Logger, l1ChainI
...
 			}
 		}
 	}
 
+	go func() {
+		if n.isIPSet {
+			return
+		}
+		updateLocalNodeTicker := time.NewTicker(initLocalNodeAddrInterval)
+		initialized := false
```
syntrust, line 298: Is this flag necessary?

ethstorage/p2p/node.go
```diff
@@ -32,6 +33,7 @@ type NodeP2P struct {
 	gater   ConnectionGater     // p2p gater, to ban/unban peers with, may be nil even with p2p enabled
 	connMgr connmgr.ConnManager // p2p conn manager, to keep a reliable number of peers, may be nil even with p2p enabled
 	// the below components are all optional, and may be nil. They require the host to not be nil.
```
syntrust, line 35: This comment can be moved down by 1 line.

ethstorage/p2p/discovery.go
ping-ke: yes, it used to change the ticker interval. if the IP has not been set yet, it will check every 30 seconds, if the IP has been updated, it will check the IP every 10 minutes

ethstorage/p2p/discovery.go
ping-ke: we can add API to query the peer ENR or out the ENR to metrics later.

syntrust (approved):

PR #218, opened 2024-02-28
Upgrade the optimism sdk version to v1.7.0

iteyelmp:
To support blobs of the Optimism type, it is necessary to upgrade the ethereum-optimism SDK to version v1.7.0, and 
 which requires synchronously upgrading the go-ethereum version. The new version has changes in the directory structure of the ethereum-optimism package and in the usage of logs in go-ethereum, necessitating adaptation.


go.mod
```diff
@@ -169,12 +170,14 @@ require (
...
+	golang.org/x/net v0.19.0 // indirect
 	google.golang.org/protobuf v1.31.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/blake3 v1.2.1 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
 
 replace github.com/iden3/go-iden3-crypto => github.com/ethstorage/go-iden3-crypto v0.0.0-20230406080944-d89aec086425
+
+replace github.com/ethereum/go-ethereum v1.13.8 => github.com/ethereum-optimism/op-geth v1.101308.2-rc.2
```
ping-ke, line 183: what is the difference between these two versions?

iteyelmp: https://github.com/ethereum-optimism/optimism/issues/9622 
They added some of their own code to the client. If it is not integrated, the compilation will prompt that the content cannot be found.

qzhodl:
I am a little concerned about our dependency on op-geth, as it may become a blocker if we need an update on geth and op-geth is not updated accordingly. Therefore, I suggest we only borrow the BLOB decoding code snippet.

PR #228, opened 2024-03-06
replace zkey

syntrust:
For public testnet 1：
1. new zkey with trust setup
2. prompt version
3. update data gas on Sepolia
4. let go of estimate gas because blobhash fails to be estimated
5. debug msg

### DIFF
--- a/cmd/es-utils/utils/utils.go
+++ b/cmd/es-utils/utils/utils.go
@@ -97,6 +97,7 @@ func SendBlobTx(
 		}
 		nonce = int64(pendingNonce)
 	}
+	log.Info("SendBlobTx", "nonce", nonce)
 
 	var gasPrice256 *uint256.Int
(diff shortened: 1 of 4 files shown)


cmd/es-utils/utils/utils.go
```diff
@@ -109,6 +110,8 @@ func SendBlobTx(
 		if nok {
 			log.Crit("Gas price is too high!", "value", val.String())
 		}
+		log.Info("SendBlobTx", "gasPriceSuggested", gasPrice256)
+		gasPrice256 = gasPrice256.Mul(gasPrice256, uint256.NewInt(2))
```
qzhodl, line 114: Do we want this logic to be checked in?

qzhodl (approved):

ping-ke (approved):

Issue #220, opened 2024-02-28
es-node should have continuous error outputs when have submitting proofing failed


qzhodl:
#### System information

EthStorage node version: `es-node version`
Build date: e.g. Tue 12 Dec 2023 04:20:27 AM EST
System version: e.g. amd64/linux
Golang version: e.g. go1.20
Commit hash : (if `develop`)

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

PR #230, opened 2024-03-09
mining stats

syntrust:
To address https://github.com/ethstorage/es-node/issues/220


ethstorage/miner/worker.go
```diff
@@ -298,6 +298,15 @@ func (w *worker) notifyResultLoop() {
...
 func (w *worker) resultLoop() {
 	defer w.wg.Done()
+	type miningError struct {
+		shardIdx uint64
+		block    *big.Int
+		err      error
+	}
+	errorCache := make([]miningError, 0)
+	totalSubmitted := 0
+	ticker := time.NewTicker(1 * time.Minute)
```
qzhodl, line 308: What is the p2p log printing internal? We may want to keep the same

qzhodl (approved):

ethstorage/miner/worker.go
```diff
@@ -312,7 +321,9 @@ func (w *worker) resultLoop() {
 				*result,
 				w.config,
 			)
+			totalSubmitted++
 			if err != nil {
+				errorCache = append(errorCache, miningError{result.startShardId, result.blockNumber, err})
```
ping-ke, line 326: this only contains L1API submit result error, may need to include other errors during mining/generating results.

qzhodl (approved):

ping-ke (approved):

PR #234, opened 2024-03-09
Rpc override

syntrust:
Dockerfile changed to support bash. Otherwise run.sh will report the error "bad substitution" in the logic of evaluating a flag.


qzhodl (approved):
LGTM

Issue #237, opened 2024-03-13
es-node cannot stop by ctrl + c with version v0.1.8


ping-ke:
#### System information

EthStorage node version: `es-node v0.1.8`
System version: e.g. amd64/ubuntu
Golang version: e.g. go1.20
Commit hash : (if `develop`)

#### Expected behaviour
After ctrl + c, the es-node should output "Starting storage node" and stop in a few seconds. 

#### Actual behaviour
After ctrl + c, the es-node outputs "Starting storage node" but the es-node still running (miner, p2p, downloader). 

#### Steps to reproduce the behaviour
5 out of 6 es-nodes with v0.1.8 cannot stop on our sepolia testnet

PR #254, opened 2024-03-23
fix issue: es-node cannot stop by ctrl + c

ping-ke:
issue https://github.com/ethstorage/es-node/issues/237

### Expected behaviour
After ctrl + c, the es-node should output "Storage node exited" and stop in a few seconds.

### Reason
The es-node cannot stop because "host" in NodeP2P failed to stop, that is due to StopNotify in host.Close() blocked by failed to get s.notifs.Lock(). the lock (read lock) is held by Network.Notify() function. And RequestShardList() in the add connection notify function would wait and block the lock.

github.com\libp2p\go-libp2p\p2p\net\swarm\swarm.go
// StopNotify unregisters Notifiee fromr receiving signals
func (s *Swarm) StopNotify(f network.Notifiee) {
	s.notifs.Lock()
	delete(s.notifs.m, f)
	s.notifs.Unlock()
}

github.com/ethstorage/es-node/ethstorage/p2p/node.go
n.host.Network().Notify(
       ... ...
       RequestShardList()
       ... ...
)

### How to fix
Move add/remove peers operations to a separate thread, and the notify function just add new peers to queue.

### How to test
Run es-node for some time, and then ctrl+c, the es-node should finish in a few seconds. 
env ES_NODE_STORAGE_MINER=<miner address> ES_NODE_SIGNER_PRIVATE_KEY=<signer> ./run.sh

### DIFF
--- a/ethstorage/p2p/node.go
+++ b/ethstorage/p2p/node.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethdb"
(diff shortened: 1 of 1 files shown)


qzhodl:
> After ctrl + c, the es-node should output "Starting storage node" and stop in a few seconds.

Why "Starting storage node" after ctrl + c?

PR #255, opened 2024-03-23
fix issue: inscrease max peers count and use p2p.peers.hi as maxPeers instead

ping-ke:
As we will have more than 100 nodes in the public testnet, we can increase the maxPeers to a larger number like 70 (prysm set this number to 70). 
As the code uses p2p.peers.lo and p2p.peers.hi to set up libp2p connection manager, so when peers count larger than p2p.peers.hi, some connections (peer count - p2p.peers.lo) would disconnect at the same time.

### How to test
run es-node to join the public testnet, and the peer count joins the local node should not be larger than 70, and when the peer count reaches 70, multi-connections (peer count - p2p.peers.lo) disconnect in a short time.

### DIFF
--- a/ethstorage/p2p/cli/load_config.go
+++ b/ethstorage/p2p/cli/load_config.go
@@ -366,10 +366,12 @@ func loadSyncerParams(conf *p2p.Config, ctx *cli.Context) error {
 	maxRequestSize := ctx.GlobalUint64(flags.MaxRequestSize.Name)
 	syncConcurrency := ctx.GlobalUint64(flags.SyncConcurrency.Name)
 	fillEmptyConcurrency := ctx.GlobalInt(flags.FillEmptyConcurrency.Name)
+	maxPeers := ctx.GlobalInt(flags.PeersLo.Name)
 	if syncConcurrency < 1 {
 		return fmt.Errorf("p2p.sync.concurrency param is invalid: the value should larger than 0")
 	}
 	conf.SyncParams = &protocol.SyncerParams{
+		MaxPeers:              maxPeers,
 		MaxRequestSize:        maxRequestSize,
 		SyncConcurrency:       syncConcurrency,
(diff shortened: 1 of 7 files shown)


qzhodl:
> As we will have more than 100 nodes in the public testnet, we can increase the maxPeers to a larger number like 60. As the code uses p2p.peers.lo and p2p.peers.hi to set up libp2p connection manager, so when peers count larger than p2p.peers.lo , some connections (peer count - p2p.peers.lo) would disconnect in the same time.
> 
> ### How to test
> run es-node to join the public testnet, and the peer count joins the local node should not be larger than 60, and no multi-connections disconnect in a short time.

I think the right description is "when peers count larger than p2p.peers.hi , some connections (peer count - p2p.peers.lo) would disconnect in the same time"?

ethstorage/p2p/cli/load_config.go
```diff
@@ -366,10 +366,12 @@ func loadSyncerParams(conf *p2p.Config, ctx *cli.Context) error {
 	maxRequestSize := ctx.GlobalUint64(flags.MaxRequestSize.Name)
 	syncConcurrency := ctx.GlobalUint64(flags.SyncConcurrency.Name)
 	fillEmptyConcurrency := ctx.GlobalInt(flags.FillEmptyConcurrency.Name)
+	maxPeers := ctx.GlobalInt(flags.PeersLo.Name)
```
qzhodl, line 369: I think there should be a comment to note that since PeersLo is set to the maxPeers here, the PeersHi param is useless

syntrust: It sounds unnatural to me to set p2p.peers.lo as maxPeers.

qzhodl (approved):
lgtm

syntrust (approved):

PR #257, opened 2024-03-25
fix issue: es-node cannot stop by ctrl + c

ping-ke:
issue https://github.com/ethstorage/es-node/issues/237

### Expected behaviour
After ctrl + c, the es-node should output "Storage node exited" and stop in a few seconds.

### Reason
The es-node cannot stop because "host" in NodeP2P failed to stop, that is due to StopNotify in host.Close() blocked by failed to get s.notifs.Lock(). the lock (read lock) is held by notifyAll which calls all the function register through Network.Notify(). And RequestShardList() in the add connection notify function would wait and block the lock, and that is because n.Host().NewStream in that function does not have timeout control. 

github.com\libp2p\go-libp2p\p2p\net\swarm\swarm.go
// StopNotify unregisters Notifiee fromr receiving signals
func (s *Swarm) StopNotify(f network.Notifiee) {
s.notifs.Lock()
delete(s.notifs.m, f)
s.notifs.Unlock()
}

github.com/ethstorage/es-node/ethstorage/p2p/node.go
n.host.Network().Notify(
... ...
RequestShardList()
... ...
)

func (n *NodeP2P) RequestShardList(remotePeer peer.ID) ([]*protocol.ContractShards, error) {
	... ...
	s, err := n.Host().NewStream(context.Background(), remotePeer, protocol.RequestShardList)
        ... ...
}

How to fix
Add timeout control to n.Host().NewStream function.

How to test
Run es-node for some time, and then ctrl+c, the es-node should finish in a few seconds.
env ES_NODE_STORAGE_MINER= ES_NODE_SIGNER_PRIVATE_KEY= ./run.sh

### DIFF
--- a/ethstorage/p2p/node.go
+++ b/ethstorage/p2p/node.go
@@ -4,10 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net"
-	"strconv"
-	"time"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/event"
@@ -26,6 +22,8 @@ import (
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 	ma "github.com/multiformats/go-multiaddr"
+	"net"
+	"strconv"
 )
 
 // NodeP2P is a p2p node, which can be used to gossip messages.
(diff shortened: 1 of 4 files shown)


ethstorage/p2p/node.go
```diff
@@ -193,7 +194,8 @@ func (n *NodeP2P) RequestL2Range(ctx context.Context, start, end uint64) (uint64
 // RequestShardList fetches shard list from remote peer
 func (n *NodeP2P) RequestShardList(remotePeer peer.ID) ([]*protocol.ContractShards, error) {
 	remoteShardList := make([]*protocol.ContractShards, 0)
-	s, err := n.Host().NewStream(context.Background(), remotePeer, protocol.RequestShardList)
+	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
```
qzhodl, line 197: Do we want to add the timeout for all the others request?

syntrust: better handle cancel? like
```
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()
```

qzhodl (approved):

syntrust (approved):

PR #270, opened 2024-04-10
Archiver spec

syntrust:


ethstorage/archiver/spec.md
```diff
@@ -0,0 +1,118 @@
...
+func (p *KZGProver) GenerateKZGProof(data []byte, sampleIdx  uint64) ([]byte, error)
+```
+5. Retrieve:
+When the API is queried, the content of blobs is loaded to generate `kzg_commitment` and `kzg_proof` on demand and served by the server.
+
+### Considerations for  Solution 2:
+- It is assumed that the beacon API `/eth/v2/beacon/blocks/{block_id}` is always available for any blocks, and the event log data from the execution layer is always available for any blocks.
+- If the performance is proven to be a problem, we may need to add a local cache/prefetch layer to save the metadata.
+- If the blob transaction hashes are required, they can also be retrieved through standard JSON-RPC API.
+
```
qzhodl, line 118: One scenario we should consider is how an es-node will respond if it receives a request for a BLOB that is not stored by EthStorage.

ethstorage/archiver/spec.md
```diff
@@ -0,0 +1,118 @@
...
+ ```
+
+ ###  Considerations for Solution 1
+- There will be a flag to control the switch of the archive API, but this flag only controls the external reading interface,  while the downloading and syncing of archive data will always be on.
+- Blobs in the latest block are not supported to be downloaded immediately. You must wait for 2 epochs and L1 to download after the block is finalized.
+
+## Solution 2: on the fly
+
+1. From slot to EL block number:
+Consider `block_id` as the beacon chain slot number or block root, we can obtain the execution layer block number by querying `/eth/v2/beacon/blocks/{block_id}` and accessing `data.message.body.execution_payload.block_number`.
```
qzhodl, line 94: Need to confirm that the `/eth/v2/beacon/blocks/{block_id}` still has valid data after 18 days

ethstorage/archiver/spec.md
syntrust: A 404 message will be returned according to spec
```
{
  "code": 404,
  "message": "Block not found"
}
```

ethstorage/archiver/spec.md
syntrust: Tested on sepolia with a slot in which blobs are supposed to be expired: https://sepolia.beaconcha.in/slot/4610252 (contains 2 blobs)

`curl -X 'GET'   'http://88.99.30.186:3500/eth/v1/beacon/blob_sidecars/4610252'   -H 'accept: application/json' `
returns `{"data":[]}`, while `curl -X 'GET'   'http://88.99.30.186:3500/eth/v2/beacon/blocks/4610252'   -H 'accept: application/json' ` still returns `execution_payload` data including block number and kzg commitments.

qizhou: Yes, the execution payload expiration will be done by EIP-4444.  We may also need to plan this in this doc (or at least put it in consideration).

qzhodl (approved):

qizhou (approved):
LGTM

PR #275, opened 2024-04-16
Blob Archiver API

syntrust:
See [the spec](https://github.com/ethstorage/es-node/blob/main/ethstorage/archiver/spec.md) for details.

### DIFF
--- a/ethstorage/archiver/api.go
+++ b/ethstorage/archiver/api.go
@@ -8,13 +8,10 @@ import (
 	"fmt"
 	"math/big"
 	"net/http"
-	"slices"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/crypto/kzg4844"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethstorage/go-ethstorage/ethstorage"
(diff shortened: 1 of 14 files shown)


ethstorage/archiver/blob.go
```diff
@@ -0,0 +1,83 @@
...
+		"}, "
+}
+
+type alias BlobSidecar
+type blobSidecarAlias struct {
+	Index         string `json:"index"`
+	KZGCommitment string `json:"kzg_commitment"`
+	KZGProof      string `json:"kzg_proof"`
+	Blob          string `json:"blob"`
+	*alias
```
qzhodl, line 42: Why we want to introduce this field?

syntrust: removed for simplicity

qzhodl (approved):
lgtm

ethstorage/archiver/api.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, hErr
+	}
+	a.logger.Info("BeaconID to execution block number", "beaconID", id, "elBlock", elBlock)
+
+	blobsInBeacon := len(kzgCommitsAll)
+	if indices != nil {
+		for _, index := range indices {
+			if int(index) >= blobsInBeacon {
+				// beacon API will ignore invalid indices and return all blobs
+				indices = nil
```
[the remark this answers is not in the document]
syntrust: As the comment says we do this to be aligned with the beacon API. Although `newIndicesError` could be an option 
 we may consider based on later tests with op stack testnet, this is an edge case anyway.

ping-ke (approved):

Issue #203, opened 2024-01-26
Support more compact encoding method in 4844 BLOB


qzhodl:
# Rationale

Why should this feature exist?
Optimism supports a more compact encoding method [here](https://github.com/ethereum-optimism/specs/blob/main/specs/derivation.md#blob-encoding), and EthStorage should also support this encoding method.

# Implementation

Do you have ideas regarding the implementation of this feature?
Are you willing to implement this feature?

PR #277, opened 2024-04-17
Support decode optimism type blob

iteyelmp:
fix issue https://github.com/ethstorage/es-node/issues/203: Minimize and integrate blob-related code in [optimism-sdk](https://github.com/ethereum-optimism/optimism/blob/develop/op-service/eth/blob.go#L88) to decode optimism type blobs.

Since using [optimism-sdk](https://github.com/ethereum-optimism/optimism/blob/develop/op-service/eth/blob.go#L88) to decode blobs requires upgrading it to v1.6.0, and optimism-sdk depends on a customized version of go-etherum(op-geth). This upgrade will cause go-etherum to be forcibly bound to the optimism version. so the blob-related code is proposed separately.


ethstorage/node/es_api.go
```diff
@@ -29,6 +29,7 @@ type DecodeType uint64
 const (
 	RawData DecodeType = iota
 	PaddingPer31Bytes
+	OptimismCompactBlob
```
qzhodl, line 32:
```suggestion
	OptimismCompact
```

ethstorage/node/es_api.go
```diff
@@ -6,11 +6,11 @@ package node
 import (
 	"bytes"
 	"errors"
-
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/log"
+	op "github.com/ethstorage/go-ethstorage/ethstorage/encoder"
```
qzhodl, line 13: 'op' is not a good name for an encoder package.

ethstorage/encoder/optimism_encoder.go
```diff
@@ -0,0 +1,224 @@
+package encoder
```
qzhodl, line 1: I would suggest not putting those codes into the encoder folder because it is for the storage encoding, not for the blob encoding. They should be together with "utils.DecodeBlob"

qzhodl (approved):

ping-ke (approved):

PR #286, opened 2024-05-16
Miner optimization - reduce RPC queries when submitting a transaction

syntrust:
The motivation of this PR is to optimize the process of mining transaction submission, reducing network interactions to save time and reduce the chance of failure caused by bad network condition.

What's in this PR:

1. Query networkID at the beginning instead of query each time when sending tx, which is not necessary.
2. Use `miningReward` of storage contract instead of the implementation in es-node to save more network communications and duplicated code.

Test can be done using a mining test case such as 
```
env ES_NODE_SIGNER_PRIVATE_KEY=<pk> ./integration_tests/run_tests.sh
```
The result shows the fix reduced the RPC queries by 4 during a mining tx submission.


qzhodl:
> 1. Save networkID at beginning
> 2. Use `miningReward` of storage contract instead of es-node.
> 
> Reduced the RPC queries by 4 during a mining tx submission.

I think two things should be included in the description of this PR:
 - The motivation of reducing the RPC queries, e.g. the network id
 - The test plan: did you finish the test, and what's the test result

qzhodl (approved):

Issue #232, opened 2024-03-09
Use dynamically adjusted request size to fetch blobs data from peers through p2p instread of using the static p2p.max.request.size value


ping-ke:
## Why should this feature exist?
Currently, we use p2p.max.request.size to control the size of blobs fetched from peers. However, different peers are distributed in different regions, so the request sizes between the local node and peers should be different. So we should use different request sizes to fetch blobs from different peers.

## How does it work?
When the local node is connected to a new peer, it still uses the p2p.max.request.size to control the size fetch blobs, if the request fails 5 times with the following error (means request timeout between the local node and the new peer), reduce the request size to half of the previous value.
t=2024-03-09T02:03:32+0000 lvl=warn msg="Failed to request blobs"               peer=16Uiu2HAmCsJgtzAZLtJPPmaSz46sasCDy7Y3xByb8NUZ1anDraCU err="i/o deadline reached"

PR #287, opened 2024-05-18
fix Issue232: Use dynamically adjusted request size to fetch blobs data from peers through p2p instread of using the static p2p.max.request.size value

ping-ke:
Issue: https://github.com/ethstorage/es-node/issues/232

Currently, we use p2p.max.request.size to control the size of blobs fetched from peers. However, different peers are distributed in different regions, so the request sizes between the local node and peers should be different. So we should use different request sizes to fetch blobs from different peers.

1. Change the p2p.max.request.size to p2p.request.size which will be used to initialize the request size from a new peer.
2. Add a tracker for each peer to adjust request size according to network conditions between the peer and local node.
	capacity = 0.9*(t.capacity) + 0.1* return blobs/second (Return blob count/Return time in second)
3. When selecting an idle peer to send a new request, order idle peers by their capacities, and select the peer with the biggest capacity.

This design refers to rates design in "github.com\ethereum\go-ethereum\eth\protocols\snap\sync.go" as the following:

A similar design also exists in prysm, it uses both capacity and score (processedBatches * 0.1) to sort, and also adds some randomness to select that peer or not.

**How To Test**
Change to the log level to debug, or change the following code in the tracker.go to info. Then run the es-node from the begining.
`log.Debug("Update tracker", "peer id", t.peerID, "elapsed", elapsed, "items", items, "old capacity", oldcap, "capacity", t.capacity)`

then check the log like the following:
`t=2024-05-25T11:46:19+0000 lvl=info msg="Update tracker"                         "peer id"=16Uiu2HAmGAyykt2njnJYTSU9KsiFutrQKZA1w8LhS45ERpqxfwFV elapsed=333.787809ms items=8,388,608 "old capacity"=39724207.481  capacity=38264942.629`

The following is the test result for this feature between local node and peer on AX101. The inital request size is 8M, and after 3 minutes, the request size stable between 4.5M ~ 5.2M
80272674f2478a3e226cc65f3c41536

### DIFF
--- a/ethstorage/flags/p2p_flags.go
+++ b/ethstorage/flags/p2p_flags.go
@@ -155,8 +155,8 @@ var (
 	InitRequestSize = cli.Uint64Flag{
 		Name: "p2p.request.size",
 		Usage: "p2p request size is the initial number of bytes to request from a remote peer. The default value is 1 * 1024 * 1024. " +
-			"It is value should not larger than 8 * 1024 * 1024. The request size will be changed according to network condition between the remote peer" +
-			"and the localnode, you can increase the max request size to improve the sync performance.",
+			"It is value should not larger than 8 * 1024 * 1024. The request size will be changed according to network condition " +
+			"between the remote peer and the local node.",
 		Required: false,
 		Value:    1 * 1024 * 1024,
 		EnvVar:   p2pEnv("MAX_REQUEST_SIZE"),
--- a/ethstorage/flags/p2p_flags.go
+++ b/ethstorage/flags/p2p_flags.go
@@ -153,10 +153,10 @@ var (
 		EnvVar:   p2pEnv("SECURITY"),
 	}
 	MaxRequestSize = cli.Uint64Flag{
-		Name: "p2p.max.request.size",
-		Usage: "max request size is the maximum number of bytes to request from a remote peer. The default value is 1 * 1024 * 1024. " +
-			"It is value should not larger than 8 * 1024 * 1024. if you have good network condition, " +
-			"you can increase the max request size to improve the sync performance.",
+		Name: "p2p.request.size",
+		Usage: "p2p request size is the initial number of bytes to request from a remote peer. The default value is 1 * 1024 * 1024. " +
+			"It is value should not larger than 8 * 1024 * 1024. The request size will be changed according to network condition between the remote peer" +
+			"and the localnode, you can increase the max request size to improve the sync performance.",
 		Required: false,
 		Value:    1 * 1024 * 1024,
 		EnvVar:   p2pEnv("MAX_REQUEST_SIZE"),
--- a/ethstorage/flags/p2p_flags.go
+++ b/ethstorage/flags/p2p_flags.go
@@ -152,7 +152,7 @@ var (
 		Value:    "noise",
 		EnvVar:   p2pEnv("SECURITY"),
 	}
-	MaxRequestSize = cli.Uint64Flag{
+	InitRequestSize = cli.Uint64Flag{
 		Name: "p2p.request.size",
 		Usage: "p2p request size is the initial number of bytes to request from a remote peer. The default value is 1 * 1024 * 1024. " +
 			"It is value should not larger than 8 * 1024 * 1024. The request size will be changed according to network condition between the remote peer" +
@@ -345,7 +345,7 @@ var p2pFlags = []cli.Flag{
 	StaticPeers,
 	HostMux,
 	HostSecurity,
-	MaxRequestSize,
+	InitRequestSize,
 	SyncConcurrency,
 	FillEmptyConcurrency,
 	MetaDownloadBatchSize,
--- a/ethstorage/p2p/protocol/peer.go
+++ b/ethstorage/p2p/protocol/peer.go
@@ -77,7 +77,7 @@ func (p *Peer) Log() log.Logger {
 	return p.logger
 }
 
-func (p *Peer) getReqestSize() uint64 {
+func (p *Peer) getRequestSize() uint64 {
 	return uint64(p.tracker.Capacity(expectRequestTime))
 }
 
@@ -100,7 +100,7 @@ func (p *Peer) RequestBlobsByRange(id uint64, contract common.Address, shardId u
 		}
 	}()
 
-	requestSize := p.getReqestSize()
+	requestSize := p.getRequestSize()
 	return SendRPC(stream, &GetBlobsByRangePacket{
 		ID:       id,
 		Contract: contract,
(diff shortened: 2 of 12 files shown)


qzhodl:
> Issue: #232
> 
> Currently, we use p2p.max.request.size to control the size of blobs fetched from peers. However, different peers are distributed in different regions, so the request sizes between the local node and peers should be different. So we should use different request sizes to fetch blobs from different peers.
> 
> 1. Change the p2p.max.request.size to p2p.request.size which will be used to initialize the request size from a new peer.
> 2. Add a tracker for each peer to adjust request size according to network conditions between the peer and local node.
>    capacity = 0.9*(t.capacity) + 0.1* return blobs/second (Return blob count/Return time in second)
> 3. When selecting an idle peer to send a new request, order idle peers by their capacities, and select the peer with the biggest capacity.

It would be great to provide the reference code link for Geth so that reviewers can compare it. Additionally, as mentioned in the ACD meeting, it would be helpful to list how Prysm implements it as well.

ethstorage/flags/p2p_flags.go
```diff
@@ -152,11 +152,11 @@ var (
 		Value:    "noise",
 		EnvVar:   p2pEnv("SECURITY"),
 	}
-	MaxRequestSize = cli.Uint64Flag{
-		Name: "p2p.max.request.size",
-		Usage: "max request size is the maximum number of bytes to request from a remote peer. The default value is 1 * 1024 * 1024. " +
-			"It is value should not larger than 8 * 1024 * 1024. if you have good network condition, " +
-			"you can increase the max request size to improve the sync performance.",
+	InitRequestSize = cli.Uint64Flag{
```
qzhodl, line 155: docs.ethstorage.io should also be updated in several places

ethstorage/p2p/protocol/peer.go
```diff
@@ -21,13 +21,15 @@ type Peer struct {
...
+	tracker     *Tracker
 	resCtx      context.Context
 	resCancel   context.CancelFunc
 	logger      log.Logger // Contextual logger with the peer id injected
 }
 
 // NewPeer create a wrapper for a network connection and negotiated  protocol version.
-func NewPeer(version uint, chainId *big.Int, peerId peer.ID, newStream newStreamFn, direction network.Direction, shards map[common.Address][]uint64) *Peer {
+func NewPeer(version uint, chainId *big.Int, peerId peer.ID, newStream newStreamFn, direction network.Direction,
+	initRequestSize float64, shards map[common.Address][]uint64) *Peer {
```
qzhodl, line 32: Is the initRequestSize value divided by blobSize? It is a little confusing

qzhodl:
What about Prysm?

ethstorage/p2p/protocol/peer.go
```diff
@@ -21,13 +21,15 @@ type Peer struct {
 	direction   network.Direction
 	version     uint                        // Protocol version negotiated
 	shards      map[common.Address][]uint64 // shards of this node support
+	tracker     *Tracker
```
qzhodl, line 24: Geth snap sync uses Trackers to calculate the mean value of all existing peers as the initial value for a newly joined peer. Do we want to follow the same practice?

qzhodl:
I feel the PR comment needs to describe how we tested it.

ethstorage/p2p/protocol/peer.go
ping-ke: I don't think this is very necessary, because it is just an initial value, and it will be updated to the appropriate value soon. Run the es-node on AX101 using the 4M as p2p.request.size and the peer's capacity will be greater than our maximum value of 8M. 

28849da6e66d19040145dfc061b1d84
t.capacity = (1-measurementImpact)*(t.capacity) + measurementImpact*measured
these 2 peers have it capacity larger than 16M and 32M, 
If p2p.request.size = 4M  (capacity = 4M/ 8s = 0.5M)
    16 * 0.1 + 0.5 * 0.9 > 1M
If p2p.request.size = 1M  (capacity = 4M/ 8s = 0.125M)
    16 * 0.1 + 0.125 * 0.9 > 1M
For peers with good network conditions, the initial value does not have much impact.

For peers with bad network conditions, each timeout request reduces capacity to 90% of the previous one.

So, change the run.sh by removing p2p.request.size setting, and the initial request size changed to 1M.

ethstorage/p2p/protocol/tracker.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Return an overestimation to force the peer out of a stuck minima, adding
+	// +1 in case the item count is too low for the overestimator to dent
+	return roundCapacity(1 + capacityOverestimation*throughput)
+}
+
+// roundCapacity gives the integer value of a capacity.
+// The result fits int32, and is guaranteed to be positive.
+func roundCapacity(cap float64) int {
+	return int(math.Min(maxRequestSize, math.Max(128*1024, math.Ceil(cap))))
```
qzhodl, line 84: blob size is 131072

qzhodl:
> Issue: #232
> 
> Currently, we use p2p.max.request.size to control the size of blobs fetched from peers. However, different peers are distributed in different regions, so the request sizes between the local node and peers should be different. So we should use different request sizes to fetch blobs from different peers.
> 
> 1. Change the p2p.max.request.size to p2p.request.size which will be used to initialize the request size from a new peer.
> 2. Add a tracker for each peer to adjust request size according to network conditions between the peer and local node.
>    capacity = 0.9*(t.capacity) + 0.1* return blobs/second (Return blob count/Return time in second)
> 3. When selecting an idle peer to send a new request, order idle peers by their capacities, and select the peer with the biggest capacity.
> 
> This design refers to rates design in "github.com\ethereum\go-ethereum\eth\protocols\snap\sync.go" as the following: 
> 
> A similar design also exists in prysm, it uses both capacity and score (processedBatches * 0.1) to sort, and also adds some randomness to select that peer or not.     
> 
> **How To Test** Change to the log level to debug, or change the following code in the tracker.go to info. Then run the es-node from the begining. `log.Debug("Update tracker", "peer id", t.peerID, "elapsed", elapsed, "items", items, "old capacity", oldcap, "capacity", t.capacity)`
> 
> then check the log like the following: `t=2024-05-25T11:46:19+0000 lvl=info msg="Update tracker" "peer id"=16Uiu2HAmGAyykt2njnJYTSU9KsiFutrQKZA1w8LhS45ERpqxfwFV elapsed=333.787809ms items=8,388,608 "old capacity"=39724207.481 capacity=38264942.629`

I think a valid test result would be: if the two nodes have a poor internet connection (e.g., China and ax101), the tracker’s capacity would quickly adapt to a very small value, and vice versa.

ethstorage/p2p/protocol/tracker.go
```diff
@@ -46,16 +46,18 @@ type Tracker struct {
 	// from disk, which is linear in the number of items, but mostly constant
 	// in their sizes.
 	peerID   string
-	capacity float64 // TODO backward compatible, change to blob count in the next testnet
+	capacity float64
+	minValue float64
```
qzhodl, line 50: It is confusing that capacity is the speed, but the minValue is actually a throughput value (speed * time). I suggest moving the minValue comparison to outsider logic (`getRequestSize`) because the tracker should not care about that.

syntrust (approved):

Issue #262, opened 2024-04-02
es-node should disconnect the connection when request fail with err "failed to dial 16Uiu2HAkumxtHzvMxsvk8QspLbWNoX1g5KE588G8FoPBckJjK88R: no addresses"


ping-ke:
#### System information

EthStorage node version: `es-node version`
System version: e.g. amd64/linux
Golang version: e.g. go1.20
Commit hash : (if `develop`)

#### Actual behaviour
Error log:
{"log":"t=2024-03-28T18:04:09+0000 lvl=warn msg=\"Failed to request blobs\"               peer=16Uiu2HAmQxR3VH1yxrnhp6spc3zLvQ7uErSD175LgbYkLQPATjUw err=\"failed to dial: failed to dial 16Uiu2HAmQxR3VH1yxrnhp6spc3zLvQ7uErSD175LgbYkLQPATjUw: no addresses\"\n","stream":"stdout","time":"2024-03-28T18:04:09.780926842Z"}
{"log":"t=2024-03-28T18:04:09+0000 lvl=warn msg=\"Failed to request blobs\"               peer=16Uiu2HAkumxtHzvMxsvk8QspLbWNoX1g5KE588G8FoPBckJjK88R err=\"failed to dial: failed to dial 16Uiu2HAkumxtHzvMxsvk8QspLbWNoX1g5KE588G8FoPBckJjK88R: no addresses\"\n","stream":"stdout","time":"2024-03-28T18:04:09.781170873Z"}
{"log":"t=2024-03-28T18:04:09+0000 lvl=warn msg=\"Failed to request blobs\"               peer=16Uiu2HAmKYpEnrkxtqqwgWMskra73hsEU552n4BdEmVTQvvQLL8L err=\"failed to dial: failed to dial 16Uiu2HAmKYpEnrkxtqqwgWMskra73hsEU552n4BdEmVTQvvQLL8L: no addresses\"\n","stream":"stdout","time":"2024-03-28T18:04:09.781409375Z"}
[...]

PR #288, opened 2024-05-22
fix issue 262: close peers that have no addresses in the host.peerstore due to expired ttl.

ping-ke:
issue 262: https://github.com/ethstorage/es-node/issues/262

**Root Cause**
When creating a newStream to send a request to a peer, if the connection does not exist, it will call the peer again. But if the address of that peer does not exist, this issue will be thrown.
The peer address does not exist because of time-to-live timeout, which is 24 hours correctly. time-to-live timeout means the peer not be added back to peerstore by discovery.

pstore.AddAddrs(info.ID, info.Addrs, discoveredAddrTTL) // discoveredAddrTTL = time.Hour * 24
 

**How to fix it**
Change the log level to debug for the error caused by newStream "no addresses" and close the peer.

### DIFF
--- a/ethstorage/p2p/node.go
+++ b/ethstorage/p2p/node.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethdb"
@@ -40,6 +41,7 @@ type NodeP2P struct {
 	syncCl         *protocol.SyncClient
 	syncSrv        *protocol.SyncServer
 	storageManager *ethstorage.StorageManager
+	resCtx         context.Context
 }
 
 // NewNodeP2P creates a new p2p node, and returns a reference to it. If the p2p is disabled, it returns nil.
@@ -67,6 +69,7 @@ func (n *NodeP2P) init(resourcesCtx context.Context, rollupCfg *rollup.EsConfig,
 	storageManager *ethstorage.StorageManager, db ethdb.Database, m metrics.Metricer, feed *event.Feed) error {
 	bwc := p2pmetrics.NewBandwidthCounter()
 	n.storageManager = storageManager
+	n.resCtx = resourcesCtx
 
 	var err error
 	// nil if disabled.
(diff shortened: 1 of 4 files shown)


qzhodl:
> the peer will remove or recover after some time

Why is the peer removed or recovered after some time? Could you please provide more explanation?

qzhodl:
I am still a little concerned because we haven’t identified the root cause of addresses disappearing in the peer store. If we change this error to the debug level, we will essentially lose visibility of the issue.

qzhodl:
> Change the log level to debug for the error caused by newStream, this should be a temporary issue between the local node and the remote peer, I have seen this issue twice (one reported by a customer and one by me - the network condition at that time was very bad.)
> 
> This issue will be resolved some times later when
> 
> 1. When the peer is removed
> 2. Addr is added back to Peerstore in the discovery.go

According to the discussion in today’s meeting, I think we may need to set up a test environment to determine if the root cause is an outdated TTL.

ping-ke:
Added log to removeExpired in addr_book.go, and will see it clear the addresses after expired. and connect to this peer will throw 
"no addresses" issue.

qzhodl:
> Added log to removeExpired in addr_book.go, and will see it clear the addresses after expired. and connect to this peer will throw "no addresses" issue.
> 
> 

Then should we also need to remove the peer in syncClient when the peerstore removes the addr?

ethstorage/p2p/protocol/syncclient.go
```diff
@@ -731,7 +732,10 @@ func (s *SyncClient) assignBlobRangeTasks() {
 
 				if err != nil {
```
syntrust, line 733: Should the error be processed immediately after `err` is got?

ethstorage/p2p/protocol/syncclient.go
```diff
@@ -1260,6 +1267,40 @@ func (s *SyncClient) ReportPeerSummary() {
 	}
 }
 
+func (s *SyncClient) PurgeBadPeers(closePeerFunc func(peer.ID) error) {
```
syntrust, line 1270: Why the peer cannot be closed by `RemovePeer()`? And does this operation need a metrics update?

ping-ke: RemovePeer is the function that removes a peer from sync client, and closePeerFunc will close the peer connection, and the RemovePeer() will be called after the connection closed.

ethstorage/p2p/protocol/syncclient.go
```diff
@@ -1260,6 +1267,40 @@ func (s *SyncClient) ReportPeerSummary() {
...
+			}
+		case <-s.resCtx.Done():
+			log.Info("P2P PurgeBadPeers stop")
+			return
+		}
+	}
+}
+
+// TODO we can add more rules for bad peers
+func (s *SyncClient) getBadPeers() []peer.ID {
```
qzhodl, line 1292: The PR description needs to be updated if the changes are more than just lowering the log level.

ethstorage/p2p/protocol/syncclient.go
```diff
@@ -731,7 +732,10 @@ func (s *SyncClient) assignBlobRangeTasks() {
 
 				if err != nil {
 					if e, ok := err.(*yamux.Error); ok && e.Timeout() {
-						log.Debug("Failed to request blobs", "peer", pr.id.String(), "err", err)
+						log.Debug("Request blobs timeout", "peer", pr.id.String(), "err", err)
+					} else if returnCode == streamError && strings.Contains(err.Error(), "no addresses") {
+						pr.isBad = true
```
qzhodl, line 737: Does Prysm have a similar issue to ours, and how do they handle it?

ping-ke: Prysm does not add addresses to peerStore, but directly connect to peer using AddrInfo.
a343d35f97d364847231b9798a7e4cb
6710635e392f82458cdc41cb35fe7bd

ethstorage/p2p/node.go
```diff
@@ -182,10 +185,36 @@ func (n *NodeP2P) init(resourcesCtx context.Context, rollupCfg *rollup.EsConfig,
 		if m != nil {
 			go m.RecordBandwidth(resourcesCtx, bwc)
 		}
+
+		go n.PurgeBadPeers()
 	}
 	return nil
 }
 
+func (n *NodeP2P) PurgeBadPeers() {
```
qzhodl, line 194: Need a comment to explain why the addresses might disappear

qzhodl (approved):

syntrust (approved):

PR #294, opened 2024-06-13
fix unit test bug: TestSaveAndLoadSyncStatus would randomly fail

ping-ke:
unit test TestSaveAndLoadSyncStatus would randomly fail as the syncCl.tasks may not sort by shard id, which is caused by the test case failure. so add sort task after tasks loaded.

### DIFF
--- a/ethstorage/p2p/protocol/syncclient.go
+++ b/ethstorage/p2p/protocol/syncclient.go
@@ -320,7 +320,7 @@ func (s *SyncClient) loadSyncStatus() {
 	}
 
 	sort.Slice(s.tasks, func(i, j int) bool {
-		return s.tasks[i].ShardId > s.tasks[j].ShardId
+		return s.tasks[i].ShardId < s.tasks[j].ShardId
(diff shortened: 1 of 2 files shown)


ethstorage/p2p/protocol/syncclient.go
```diff
@@ -318,6 +318,10 @@ func (s *SyncClient) loadSyncStatus() {
 		t := s.createTask(sid, lastKvIndex)
 		s.tasks = append(s.tasks, t)
 	}
+
+	sort.Slice(s.tasks, func(i, j int) bool {
+		return s.tasks[i].ShardId > s.tasks[j].ShardId
```
qzhodl, line 323: tasks[0]'s shardId should be 0 or 3?

ping-ke: change the order to start from 0

qzhodl (approved):

syntrust (approved):

PR #297, opened 2024-06-25
Support flag override for both init and start

syntrust:
The changes aim to support both es-node init and start in the case of flags override by appending them after `run.sh`. 
The flags that can be used this way includes common flags in both actions, which are
  --datadir
  --l1.rpc
  --storage.l1contract
  --storage.miner


syntrust:
replaced by https://github.com/ethstorage/es-node/pull/310

PR #305, opened 2024-07-04
Batch download into cache

qzhodl:
The very long finalization time on L2 (12 hours) breaks our initial assumption about `downloadToCache`. Therefore, we may need to batch download the BLOB to cache in case there are too many events emitted from the storage contract and the es-node can’t handle them all at once.

### DIFF
--- a/ethstorage/downloader/downloader.go
+++ b/ethstorage/downloader/downloader.go
@@ -227,13 +227,13 @@ func (s *Downloader) downloadToCache() {
 		}
 		_, err := s.downloadRange(start+1, rangeEnd, true)
 
-		if err == nil {
-			s.lastCacheBlock = rangeEnd
-			start = rangeEnd
-		} else {
+		if err != nil {
 			s.log.Error("DownloadRange failed", "err", err)
 			return
 		}
+
+		s.lastCacheBlock = rangeEnd
+		start = rangeEnd
 	}
 }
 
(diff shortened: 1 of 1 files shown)


iteyelmp (approved):

ethstorage/downloader/downloader.go
```diff
@@ -220,14 +220,20 @@ func (s *Downloader) downloadToCache() {
...
-	} else {
-		s.log.Info("DownloadRange failed", "err", err)
+	for start < end {
+		rangeEnd := start + downloadBatchSize
+		if rangeEnd > end {
+			rangeEnd = end
+		}
+		_, err := s.downloadRange(start+1, rangeEnd, true)
+
+		if err == nil {
```
syntrust, line 230: Anything should be done to handle the error besides just logging it? For another, would it be more natural to follow the "err!=nil" pattern?

ping-ke (approved):

ethstorage/downloader/downloader.go
ping-ke: yes, err != nil would look better.
if err != nil {
        s.log.Error("DownloadRange failed", "err", err)
	 return
}
s.lastCacheBlock = rangeEnd
start = rangeEnd

qzhodl: > Anything should be done to handle the error besides just logging it?
The error is most likely due to network conditions and RPC endpoint errors. We don’t need to retry here because after a slot, another latest block will arrive and call the `downloadToCache` function again.

syntrust (approved):

Issue #298, opened 2024-06-25
Need a disk-backed pool for non-finalized BLOBs


qzhodl:
# Rationale

**Why should this feature exist?**

Currently, the downloader uses a memory cache to keep non-finalized BLOBs and only writes them to disk after they are finalized. While this approach may be sufficient for L1, where the finalization time is usually around 12 minutes, it might not be adequate for L2. On L2, the finalization time is mostly around 12 hours, which means the memory may not be enough. Therefore, we might need a disk-backed pool for L2 to handle the longer finalization times.

**What are the use-cases?**

L2 case.

# Implementation

**Do you have ideas regarding the implementation of this feature?**

Geth implements a disk-backed transaction pool ( [billy](https://github.com/holiman/billy?tab=readme-ov-file) and limbo )

syntrust:
Done by https://github.com/ethstorage/es-node/pull/306

PR #306, opened 2024-07-05
Disk-based cache for the downloader

syntrust:
This PR is purposed to solve the issue https://github.com/ethstorage/es-node/issues/298.

To hold the downloaded blobs in es-node in a scenario like long-time finalization, the blob cache uses https://github.com/holiman/billy as the disk-backed solution. Billy is a very simplistic data store with features like:
-  It has fixed item-sized buckets, represented by flat files. 
- The payloads will be roughly equally heavy on write, delete, and read operations.
- The data is expected to be transient rather than pertinent stored.
- Upon startup, Billy compacts the files to remove any accumulated gaps.

Since blob transactions can contain between 1-6 blobs, the stored items in the blob cache here are expected to have sizes of 128KB  ~ 128KB * 6 buckets. 

A similar implementation reference is the blobpool of geth: https://github.com/ethereum/go-ethereum/pull/26940

Other changes include:
- replace hash with block number as key in blockblob cache to handle re-org
- use a customized billy implementation to support reading samples directly from the disk cache for better performance.

Tests should verify that sampling from cached blobs can work as fast as from storage file, and the mining tx confirmed successfully.


ethstorage/downloader/blob_disk_cache.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	id, err := c.store.Put(rlpBlock)
+	if err != nil {
+		c.lg.Error("Failed to write blockBlobs into storage", "block", block.number, "err", err)
+		return err
+	}
+	c.lookup[block.hash] = id
```
qzhodl, line 113: as we discussed before, we may want to use block number to index blocks

ethstorage/downloader/blob_disk_cache.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	item := new(blockBlobs)
+	if err := rlp.DecodeBytes(data, item); err != nil {
+		c.lg.Error("Failed to decode block", "id", id, "err", err)
+		return nil, err
+	}
+	return item, nil
+}
+
+func (c *BlobDiskCache) Close() error {
```
qzhodl, line 223: When we close the cache, we may want to clean the cache anyway, and the downloader will download the blobs from last saved finalized block when restart

ethstorage/downloader/downloader.go
```diff
@@ -80,13 +83,79 @@ type blob struct {
...
+
+func (b *blob) DecodeRLP(s *rlp.Stream) error {
+	var decodedData struct {
+		KvIndex *big.Int
+		KvSize  *big.Int
+		Hash    common.Hash
+		Data    []byte
+	}
+
+	if err := s.Decode(&decodedData); err != nil {
```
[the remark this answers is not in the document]
syntrust: It will stackoverflow

syntrust:
After using an extend GetSample API on billy, the performance issue disappears.

Changes to billy can be reviewed here: 
https://github.com/holiman/billy/compare/main...ethstorage:billy:main

ethstorage/downloader/blob_cache_test.go
```diff
@@ -7,39 +7,97 @@ import (
...
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto/kzg4844"
 	"github.com/ethstorage/go-ethstorage/ethstorage"
+	"github.com/ethstorage/go-ethstorage/ethstorage/log"
 	"github.com/protolambda/go-kzg/eth"
 )
 
 var (
-	bc        BlobCache
+	cache     BlobCache
```
qzhodl, line 24: This is for the unit test. I think we may need to set up a real test environment to see the results:

1.	Deploy a test storage contract with a shard size of 512 GB.
2.	Upload 3 to 6 blobs every 2 seconds.

Then, we should collect the results, including:

	1.	Ensuring that the memory usage of es-node remains stable.
	2.	The disk-backed cache should be around 16 GB (131072 * 3 * 3600 * 12) just before finalization.
	3.	Monitor how long it takes to delete all the cache items afterward.
	4.	Record the time required to write the 16 GB of data to the disk after finalization, noting any exceptions or issues during this process.
	5.	Monitor the time it takes to close the cache when shutting down es-node.

ethstorage/downloader/blob_disk_cache.go
```diff
@@ -58,12 +62,18 @@ func (c *BlobDiskCache) SetBlockBlobs(block *blockBlobs) error {
 
 	var blbs []*blob
 	for _, b := range block.blobs {
+		kvi := b.kvIndex.Uint64()
+		if id, ok := c.kvIndexLookup[kvi]; ok {
+			if err := c.store.Delete(id); err != nil {
```
qzhodl, line 67: This is not the situation I mentioned. What I referred to above concerns the re-org case. Actually, this change will cause an error because the earlier block that first updates the key-value will not retrieve its data back.

qzhodl (approved):

ethstorage/downloader/blob_cache_test.go
syntrust: will be done on L2 testnet

ping-ke (approved):

syntrust:
> This is for the unit test. I think we may need to set up a real test environment to see the results:
> 
> 1. Deploy a test storage contract with a shard size of 512 GB.
> 2. Upload 3 to 6 blobs every 2 seconds.
> 
> Then, we should collect the results, including:
> 
> 1. Ensuring that the memory usage of es-node remains stable.
> 2. The sampling time should be the same
> 3. The disk-backed cache should be around 16 GB (131072 * 3 * 3600 * 12) just before finalization.
> 4. Monitor how long it takes to delete all the cache items afterward.
> 5. Record the time required to write the 16 GB of data to the disk after finalization, noting any exceptions or issues during this process.
> 6. Monitor the time it takes to close the cache when shutting down es-node.

**Updates on performance test based on the above comments**

The test has been done with build with commit 264bf03299093f1cb2e8597bc34ec7632257598f
1. The memory usage of the blob cache remains stable. 
2. The sampling time remains unchanged despite the cache is empty or less or many.
3. Since the finalization happens frequently, disk usage changes are not noticeable.
4. The cleanup method takes around 10ms to delete 1000+ blobs
5. No exception found when finalization happens
6. It takes no time to delete the cache file.

Issue #266, opened 2024-04-04
config: support directory in miner.zkey


jhvst:
# Rationale

Why should this feature exist?

Currently, `es-node` mixes build-time artifacts (the snarkjs folder) with runtime artifacts (the zkey files). This causes an awkward situation for deterministic builds which produce a read-only folder derivation, such as with the Nix package manager. Here, the zkey files cannot be placed under the snarkjs folder unless the zkeys are downloaded during the build process. Bundling the zkeys with the package would balloon the derivation size by additional 550MB, which makes it hard to distribute.

What are the use-cases?

By allowing the zkeys to reside in some other folder, one could have e.g. a systemd script which ensures that the files are in the directory, while still using the build-time artifacts for snarkjs. 

# Implementation

Do you have ideas regarding the implementation of this feature?

Separate the `--miner.zkey` from utilizing `--miner.zk-working-dir`, and instead of a single file name, allow an absolute or relative path to be configured for it.

Are you willing to implement this feature?

No, I currently do not have the bandwidth to make the changes.

syntrust:
Hi jhvst, thanks for your suggestion. We will plan this enhancement soon.

syntrust:
Close as solved in https://github.com/ethstorage/es-node/pull/310

PR #310, opened 2024-07-19
Move init data file functionality out of run script

syntrust:
The background is that as more features are supported, the configurations become more and more complex, and maintaining the initialization and startup nodes in one script file becomes increasingly unsustainable.

So we separate the init from run so that the flags can be managed separately.

Also supported `--miner.zkey` as path to address https://github.com/ethstorage/es-node/issues/266.

Related tutorials change - https://github.com/ethstorage/ethstorage-doc/pull/32.

Tests need to be covered:
- `init.sh`: data files created and zkey downloaded, snarkjs installation check in all installation options:
  - pre-built
  - from source
  - use pre-built Docker 
  - use local Docker
    - docker-compose
    - run-docker.sh
- `run.sh` starts es-node and mining successfully with `--miner.zkey` as the file path in all the above 3 options.
  - especially in Docker, there is no need to download zkey for a new build as it is mapped to the host disk


run.sh
```diff
@@ -1,9 +1,12 @@
...
 # env ES_NODE_STORAGE_MINER=<miner> ES_NODE_SIGNER_PRIVATE_KEY=<private_key> ./run.sh
-# for one zk proof per sample (if the storage contract supports):
-# env ES_NODE_STORAGE_MINER=<miner> ES_NODE_SIGNER_PRIVATE_KEY=<private_key> ./run.sh --miner.zk-prover-mode 1 --l1.rpc <el_rpc> --l1.beacon <cl_rpc>
+# usage 2 (overriding rpc urls):
+# env ES_NODE_STORAGE_MINER=<miner> ES_NODE_SIGNER_PRIVATE_KEY=<private_key> ./run.sh --l1.rpc <el_rpc> --l1.beacon <cl_rpc>
+
+# Note: currently only zk prover mode 2 is supported
+zkey_file="./build/bin/snark_lib/zkey/blob_poseidon2.zkey"
 
 if [ -z "$ES_NODE_STORAGE_MINER" ]; then
```
qzhodl, line 11: what if we did not check these env variables and let the es-node do the check?

run.sh
```diff
@@ -25,153 +28,37 @@ if [ ${#ES_NODE_SIGNER_PRIVATE_KEY} -ne 64 ]; then
...
-if [ ! -e  ${zkey_file} ] || [ $(wc -c <  ${zkey_file}) -ne ${zkey_size} ]; then
-  echo "Start downloading ${zkey_file}..." 
-  curl $zkey_url -o ${zkey_file}
-  if [ ! -e  ${zkey_file} ]; then
-    echo "Error: The zkey file was not downloaded. Please try again."
-    exit 1
-  fi
-  if [ $(wc -c <  ${zkey_file}) -ne ${zkey_size} ]; then
-    echo "Error: The zkey file was not downloaded correctly. You can check the file content for more information."
+if [ ! -e $storage_file_0 ]; then
```
qzhodl, line 39: Do we also want to check the zkey file here?

init.sh
```diff
@@ -0,0 +1,135 @@
+#!/bin/bash
+
+# usage 1 (use go-rapisnark):
+# env ES_NODE_STORAGE_MINER=<miner> ./init.sh
+# usage 2 (use snarkjs, will check Node.js version and snarkjs installation):
+# env ES_NODE_STORAGE_MINER=<miner> ./init.sh --miner.zk-prover-impl 1
+
+if [ -z "$ES_NODE_STORAGE_MINER" ]; then
```
qzhodl, line 8: There are many same code shared between shell script, could we include the same code as a common file?

run.sh
syntrust: Will leave both checks to the node, e.g.,

```
 CRIT [08-08|10:32:24.330] Init ZK prover failed                    error="zkey does not exist" dir=/Users/dl/code/es-node/build/bin/snark_lib/zkey/blob_poseidon2.zkey
```

run-l2.sh
```diff
@@ -5,26 +5,6 @@
 
 zkey_file="./build/bin/snark_lib/zkey/blob_poseidon2.zkey"
 
-if [ -z "$ES_NODE_STORAGE_MINER" ]; then
```
qzhodl, line 8: Can we reuse the run.sh as the same way as init-l2.sh?

run-l2.sh
```diff
@@ -1,176 +1,33 @@
...
-        echo "snarkjs installed successfully."
-      else
-        echo "Error: snarkjs install failed with the following error:"
-        echo "$snarkjs_install"
-        exit 1
-      fi
-  fi
-
-fi
+zkey_file="./build/bin/snark_lib/zkey/blob_poseidon2.zkey"
```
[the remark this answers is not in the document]
ping-ke: --miner.zk-prover-impl also need to be the same between init*.sh and run*.sh, otherwise the following will fail.
```
./init.sh --miner.zk-prover-impl=2 
./run.sh 
```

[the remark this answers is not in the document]
syntrust: ≈

> this is related to --miner.zk-prover-mode, should not hard code here. otherwise, the init with --miner.zk-prover-mode=1 will failed as the init download blob_poseidon.zkey while it use blob_poseidon2.zkey.
> 
> maybe we can use init*.sh in the run*.sh

You can override the zkey file by using the `--miner.zkey` flag. Add comments in run.sh.


> --miner.zk-prover-impl also need to be the same between init*.sh and run*.sh, otherwise the following will fail.
> 
> ```
> ./init.sh --miner.zk-prover-impl=2 
> ./run.sh 
> ```

Documented in https://github.com/ethstorage/es-node/pull/310

qzhodl (approved):

ping-ke (approved):

Issue #308, opened 2024-07-15
Add integration test to action and run regularly


ping-ke:
# Rationale

We need an integration test to check if the latest code can run successfully, and we can set it up as an action to run regularly. We need the test to cover the following:
1. Contract can be deployed to l2 successfully
2. es-node bootnode can be run successfully
3. blobs can be uploaded and downloaded to es-node 
4. es-node can be run correctly
5. es-node can update its state regularly
6. es-node can sync finish within the expected time 
7. es-node can mined 

# Implementation
To run the integration test, it is better to run on a self-host machine as it needs to run for a long time with a larger storage size which will incur larger charges. 

## why self-host machine
https://github.com/pricing/calculator?feature=actions#actions
https://docs.github.com/zh/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates

## How to add self-host machine
https://docs.github.com/zh/actions/hosting-your-own-runners/managing-self-hosted-runners/adding-self-hosted-runners

## Code change
1. Create a new run-integration-test.sh 
2. Create an Integration test service that has the same RPC handlers function and receives es-node state and check states
```
	http.HandleFunc("/hello", HelloHandler)
	http.HandleFunc("/reportstate", ReportStateHandler)
```
3. Add action integration-test.yml 

## Check Points
1. Contract can be deployed successfully
2. bootnode can be run 
3. Blobs can be uploaded and downloaded to es-node successfully
4. Receive the Hello request within 10 minutes; otherwise, the es-node may be down.
5. Get states updated every 5 minutes; otherwise, the es-node may be down
6. Sync percentage should increase every state update before it reaches 100%; 
7. Get mine state after sync is done;
8. Mined at least one block within 2 hours.

Note: we can also run full node as an integration test every week or month

PR #313, opened 2024-08-05
Create action to run Integration test regularly 

ping-ke:
related issue: https://github.com/ethstorage/es-node/issues/308

We need an integration test to check if the latest code can run successfully, and we can set it up as an action to run regularly. 
## Code change
1. Create a new run-l2-it-rpc.sh and run-l2-it.sh 
2. Create an Integration test service that has the same RPC handlers function and receives es-node state and check states
```
	http.HandleFunc("/hello", HelloHandler)
	http.HandleFunc("/reportstate", ReportStateHandler)
```
3. Add action integration-test.yml 

## Check Points
1. Contract can be deployed successfully
2. bootnode can be run successfully
3. Blobs can be uploaded and downloaded to es-node successfully
4. Receive the Hello request within 10 minutes; otherwise, the es-node may be down.
5. Get states updated every 5 minutes; otherwise, the es-node may be down.
6. Sync percentage should increase every state update before it reaches 100%; 
7. Get mine state after sync is done;
8. Mined at least one block within 2 hours.
9. Verify data from RPC 
10. Verify data from shard manager

The integration test runs on a self-host machine because it needs to run for a long time with a larger storage size which will incur larger charges. 
https://github.com/pricing/calculator?feature=actions#actions
https://docs.github.com/zh/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates


.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,59 @@
...
+          ./run-l2-it-rpc.sh > es-node-it-bootnode.log&
+
+      - name: Upload Blobs
+        run: |
+          cd ../ethstorage-sdk 
+          git checkout integration-test
+          git pull
+          npm install --force
+          rollup -c
+          node ituploader.js
```
qzhodl, line 51: This script should not be in the sdk

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,59 @@
...
+          git submodule init && git submodule update
+          npx hardhat run scripts/deployL2-it.js --network qkc_testnet >> deploy.log
+          echo ES_NODE_CONTRACT_ADDRESS=`cat .caddr` >> "$GITHUB_ENV"
+          echo $ES_NODE_CONTRACT_ADDRESS
+
+      - name: Build and Run Bootnode Node
+        run: |
+          cd ../es-node
+          make
+          ./run-l2-it-rpc.sh > es-node-it-bootnode.log&
```
qzhodl, line 42: Why the bootnode is started as a RPC node?

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,59 @@
...
+          git pull
+          npm install --force
+          rollup -c
+          node ituploader.js
+
+      - name: Test
+        run: |
+          cd ../es-node
+          env ES_NODE_STORAGE_MINER=0x5C935469C5592Aeeac3372e922d9bCEabDF8830d ES_NODE_SIGNER_PRIVATE_KEY=e690c916d02b6f3ad4bb70cc2663389ea88a79ded45613b1bd5545961d063258 ./run-l2-it.sh > es-node-it.log&
+          cd ./cmd/integration-test-server
```
qzhodl, line 57: are there any success and failed tasks that we can inspect what they look like?

.github/workflows/integration-test.yml
ping-ke: the other node covers mining, syncing, and encoding/decoding parts, this bootnode can be run as RPC to cover more.

.github/workflows/integration-test.yml
ping-ke: Currently, we will check the final state after collected 2 hours node states, and failed service if any error collected.

```
	time.Sleep(executionTime)
	checkFinalState(lastRecord)

	if len(errorMessages) > 0 {
		panic(fmt.Sprintf("integration test fail %v", errorMessages))
	}
```

run-l2-it.sh
```diff
@@ -0,0 +1,187 @@
+#!/bin/bash
+
+# usage:
+# env ES_NODE_STORAGE_MINER=<miner> ES_NODE_SIGNER_PRIVATE_KEY=<private_key> ES_NODE_CONTRACT_ADDRESS=<your_contract_address> ./run-l2-it.sh
+# for one zk proof per sample (if the storage contract supports):
+# env ES_NODE_STORAGE_MINER=<miner> ES_NODE_SIGNER_PRIVATE_KEY=<private_key> ES_NODE_CONTRACT_ADDRESS=<your_contract_address> ./run-l2-it.sh --miner.zk-prover-mode 1 --l1.rpc <el_rpc> --l1.beacon <cl_rpc>
+
+if [ -z "$ES_NODE_STORAGE_MINER" ]; then
```
qzhodl, line 8: This PR may depend on this [PR](https://github.com/ethstorage/es-node/pull/310)

cmd/integration-test-server/main.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+}
+
+func addErrorMessage(errMessage string) {
+	log.Warn("Add error message", "msg", errMessage)
+	errorMessages = append(errorMessages, errMessage+"\n")
+}
+
+func listenAndServe(port int) error {
+	http.HandleFunc("/hello", HelloHandler)
```
syntrust, line 173: Who calls the 2 RPC methods?

ping-ke: es-node will call this to register to the metrics service/integration server.

cmd/integration-test-server/main.go
```diff
@@ -0,0 +1,192 @@
...
+		addErrorMessage("No state submitted during the test")
+		return
+	}
+	if !hasConnectedPeer {
+		addErrorMessage("es-node peer count should larger than 0")
+	}
+
+	log.Info("Final state", "id", state.Id, "version", state.Version)
+	for _, shardState := range state.Shards {
+		if shardState.SyncState.SyncProgress != 10000 {
```
qzhodl, line 141: Any where to check "Blobs can be downloaded to es-node successfully" && "Synced data is correct."?

integration_tests/scripts/ituploader.js
```diff
@@ -0,0 +1,72 @@
...
+                setTimeout(() => console.log("Upload done!"), 300000)
+                clearInterval(intervalId);
+            }
+        } catch (e) {
+            console.error(`EthStorage: get finalized block failed!`, e.message);
+        }
+    }, 120000);
+}
+
+UploadBlobsForIntegrationTest();
```
qzhodl, line 71: How does the github workflow know if this task success or not?

ping-ke: add the following to exit with failed.
core.setFailed(`EthStorage: get latest block failed with message: ${e.message}`);

ethstorage/p2p/node.go
```diff
@@ -96,11 +96,11 @@ func (n *NodeP2P) init(resourcesCtx context.Context, rollupCfg *rollup.EsConfig,
 					shards       map[common.Address][]uint64
 					remotePeerId = conn.RemotePeer()
 				)
-				if len(n.host.Peerstore().Addrs(remotePeerId)) == 0 {
+				if len(n.host.Peerstore().Addrs(remotePeerId)) == 0 && conn.Stat().Direction == network.DirOutbound {
```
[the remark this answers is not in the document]
ping-ke: len(n.host.Peerstore().Addrs(remotePeerId)) == 0 is just for inbound connection, make a small correction

integration_tests/scripts/ituploader.js
qzhodl: What about other places that might cause errors, such as upload failures or gas estimation failures?

integration_tests/scripts/ituploader.js
```diff
@@ -0,0 +1,75 @@
...
+const privateKey = process.env.ES_NODE_UPLOADER_PRIVATE_KEY;
+const contractAddr = process.env.ES_NODE_CONTRACT_ADDRESS;
+const RPC = 'http://65.109.20.29:8545';
+const contractABI = [
+    "function lastKvIdx() public view returns (uint40)"
+]
+
+const provider = new ethers.JsonRpcProvider(RPC);
+const contract = new Contract(contractAddr, contractABI, provider);
+const MAX_BLOB = 256n;
```
qzhodl, line 18: The test shard contains 32k blobs, and if we only upload 256 blobs, it will be very difficult during the mining process to hit the key-value pairs that contain actual data and submit a proof that includes them. This means we might miss a very important case.

ethstorage/p2p/node.go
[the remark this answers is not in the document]
qzhodl: The specific scenario is still unclear, but if it’s not related to this PR, the best practice would be to create a new PR.

cmd/integration-test-server/main.go
```diff
@@ -161,6 +180,52 @@ func checkFinalState(state *node.NodeState) {
...
+		if bytes.Compare(blobs[0][:], data) != 0 {
+			return errors.New(fmt.Sprintf("compare data %d fail, expected data %s; data: %s",
+				i, common.Bytes2Hex(blobs[0][:256]), common.Bytes2Hex(data[:256])))
+		}
+		i++
+	}
+	return nil
+}
+
+func generateMetadata(hash common.Hash) common.Hash {
```
qzhodl, line 220: Where do we use this function?

cmd/integration-test-server/main.go
```diff
@@ -0,0 +1,260 @@
...
+		checkState(lastRecord, state)
+	}
+
+	lastRecord = state
+
+	w.Write([]byte(`{"status":"ok"}`))
+}
+
+func checkState(oldState, newState *node.NodeState) {
+	if len(oldState.Shards) != len(newState.Shards) {
```
syntrust, line 107: Should we consider the case of opening new shards?

integration_tests/scripts/ituploader.js
```diff
@@ -0,0 +1,75 @@
...
+    let latestBlock
+    try {
+        latestBlock = await provider.getBlock();
+        console.log("latest block number is", latestBlock.number);
+    } catch (e) {
+        core.setFailed(`EthStorage: get latest block failed with message: ${e.message}`);
+        return
+    }
+
+    // wait for blobs finalized
```
syntrust, line 54: Why do we need to wait for the blobs to be finalized?

ping-ke: the bootnode only downloads and writes to files for finalized blobs, otherwise, those blobs only saved in memory and cannot sync to peer.

cmd/integration-test-server/main.go
ping-ke: We cannot add or remove shards while running es-node, right? We can only update the node's shards by sharding it down and restarting it again.

integration_tests/scripts/ituploader.js
ping-ke: Change to 4096, as each submission will sample 8 times, so every submission will have chances to cover blobs that are not empty.

integration_tests/scripts/ituploader.js
syntrust: Just wondering how to add the case of mining from the downloader cache...

cmd/integration-test-server/main.go
syntrust: Actually we can init the shards beforehand in the local es-node, but it depends on the requirements of the tests.

ping-ke: let's add after multi-shards support PR check-in?

integration_tests/scripts/ituploader.js
```diff
@@ -33,12 +33,22 @@ async function UploadBlobsForIntegrationTest() {
...
+        }
 
         // write
-        let status = await es.write(buf.subarray(0,32).toString('hex'), buf);
-        console.log(status);
+        try {
+            let status = await es.write(buf.subarray(0,32).toString('hex'), buf);
+            console.log(status);
+        } catch (e) {
+            console.log("upload blob error:", e.message);
```
qzhodl, line 49: Do we want to call core.setFailed?

integration_tests/scripts/ituploader.js
qzhodl: I think we should check whether we have sampled the key-value pairs that contain the actual data

cmd/integration-test-server/main.go
qzhodl: If we agree that this is the case we want to include in the integration test, I think you should add it to the checkpoints in the PR description. Also, please leave a note mentioning that we will need a separate PR to add this case if we have to wait for the multi-shard support PR to be merged

integration_tests/scripts/ituploader.js
qzhodl: Is that also a potential case we should consider adding to the integration test?

syntrust:
Not sure in this PR but, since we already have an RPC node, should we support the cases of getting blobs (of different blob encoding types) from the API as part of integration tests?

integration_tests/scripts/ituploader.js
ping-ke: this may be a temporary issue, we can let it continue uploading until it timeout

integration_tests/scripts/ituploader.js
ping-ke: after we support multi-shard, we should fill in all shard 0, so the sampling will cover the key-value pairs.

cmd/integration-test-server/main.go
ping-ke: add TODO for this.

integration_tests/scripts/ituploader.js
ping-ke: add TODO for this

integration_tests/scripts/ituploader.js
qzhodl: Why did we resolve this issue?

qzhodl:
> Not sure in this PR but, since we already have an RPC node, should we support the cases of getting blobs (of different blob encoding types) from the API as part of integration tests?

@ping-ke do we also want to add a TODO for this

qzhodl (approved):
LGTM

syntrust (approved):

PR #316, opened 2024-09-09
fix bug: connection with DirInbound may not have addresses in Peerstore.

ping-ke:
Connection is filtered out when n.host.Peerstore().Addrs(remotePeerId)) == 0. which is added to fix issue https://github.com/ethstorage/go-ethstorage/issues/72. (related PR: https://github.com/ethstorage/go-ethstorage/pull/91)
However, this fix will additionally filter out some connections that should not be ignored. For example, if a connection is coming in connection (Direction == DirInbound), its address may not have been added to Peerstore yet, at this time, the incoming connection would be ignored. The following are the connections shown when stopping the es-node, but only 29 of them are added to the syncclient.

**How to Fix**
To ignore the connection without disconnecting which is caused by NAT, this can be done after we get the shard list failure. 

After the change, we will get the following error for that case (temporary change its log level from debug to warning to make it more obvious) 

Running the change on Sepolia test net before and after the change for 10 minutes, and the connected connections increase from 29 to 40.
Before change:

After change:

### DIFF
--- a/ethstorage/p2p/node.go
+++ b/ethstorage/p2p/node.go
@@ -103,8 +103,8 @@ func (n *NodeP2P) init(resourcesCtx context.Context, rollupCfg *rollup.EsConfig,
 					// n.RequestShardList to fetch the shard list of the new node.
 					remoteShardList, e := n.RequestShardList(remotePeerId)
 					if e != nil && len(n.host.Peerstore().Addrs(remotePeerId)) == 0 {
-						// As the node host enable NATService, which will create a new connection with another
-						// peer id and its Addrs will not be set to Peerstore. So if len of peer Addrs is 0 and
+						// As the remote node host may enable NATService, which will create a new connection with another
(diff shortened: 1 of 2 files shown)


ethstorage/p2p/node.go
```diff
@@ -96,7 +96,7 @@ func (n *NodeP2P) init(resourcesCtx context.Context, rollupCfg *rollup.EsConfig,
 					shards       map[common.Address][]uint64
 					remotePeerId = conn.RemotePeer()
 				)
-				if len(n.host.Peerstore().Addrs(remotePeerId)) == 0 {
+				if len(n.host.Peerstore().Addrs(remotePeerId)) == 0 && conn.Stat().Direction == network.DirOutbound {
```
syntrust, line 99: So the consequence of this bug is we ignored some incoming connections?

ping-ke: yes

syntrust (approved):

qzhodl (approved):

PR #326, opened 2024-10-25
Add check by design failure check to integration test

ping-ke:
Currently, two failures ("diff not match" and "invalid samples") are by-design failures, which should not cause the integration test to fail. So, when the mining failure count is larger than 0, the code will check whether those failures are by design or not. 
Log errors and remove them from the failure count if they are by design.

### DIFF
--- a/cmd/integration-test-server/main.go
+++ b/cmd/integration-test-server/main.go
@@ -14,6 +14,8 @@ import (
 	"math"
 	"net/http"
 	"os"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/crate-crypto/go-proto-danksharding-crypto/eth"
@@ -41,6 +43,7 @@ const (
 	uploadedDataFile = ".data"
 	shardFile0       = "../../es-data-it/shard-0.dat"
 	shardFile1       = "../../es-data-it/shard-1.dat"
+	logFile          = "../../es-node-it.log"
 )
 
 var (
@@ -175,7 +178,11 @@ func checkFinalState(state *node.NodeState) {
 			addErrorMessage("At lease one block should be mined successfully during the test.")
 		}
 		if shardState.SubmissionState.Failed > 0 {
-			addErrorMessage(fmt.Sprintf("%d submission failed during the test.", shardState.SubmissionState.Failed))
+			byDesignCount := checkByDesignFailure()
+			failureCount := shardState.SubmissionState.Failed - byDesignCount
+			if failureCount > 0 {
+				addErrorMessage(fmt.Sprintf("%d submission failed during the test.", failureCount))
+			}
 		}
 		log.Info("Final state", "id", state.Id, "shard", shardState.ShardId, "miner", shardState.Miner, "sync progress",
 			shardState.SyncState.SyncProgress, "fill progress", shardState.SyncState.FillEmptyProgress, "mining power",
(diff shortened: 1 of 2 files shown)


cmd/integration-test-server/main.go
```diff
@@ -278,6 +288,202 @@ func downloadBlobFromRPC(client *rpc.Client, kvIndex uint64, hash common.Hash) (
...
+	for fileScanner.Scan() {
+		logText := fileScanner.Text()
+		if strings.Contains(logText, "Set mining result") {
+			block, nonce, err := fetchMinedBlockAndNonce(logText)
+			if err != nil {
+				log.Error("checkDiffNotMatchError error", "log", logText, "error", err.Error())
+				continue
+			}
+			if n, ok := miningResults[block]; ok && strings.Compare(n, nonce) != 0 {
+				conflicts[block] = true
```
qzhodl, line 332: I think this condition may be a little stricter than neccessary

cmd/integration-test-server/main.go
```diff
@@ -278,6 +288,202 @@ func downloadBlobFromRPC(client *rpc.Client, kvIndex uint64, hash common.Hash) (
...
+
+	return count0 + count1
+}
+
+func checkDiffNotMatchError() (int, error) {
+	// Description: 1. got two mining results with the same block number and diff nonce; 2. two mining results, one successful and one failed.
+	// Sample:
+	// lvl=info msg="Set mining result"         shard=1 block=6,906,682 nonce=565,740
+	// lvl=info msg="Set mining result"         shard=1 block=6,906,682 nonce=779,947
+	// lvl=eror msg="Failed to submit mined result"         shard=1 block=6,906,682 error="failed to estimate gas: execution reverted: StorageContract: diff not match"
```
syntrust, line 309: The logic should be: check the "Set mining result" timestamp of the block mined, if any mining tx confirmed since then, the "diff not match" error should be omitted.

cmd/integration-test-server/main.go
```diff
@@ -278,6 +288,202 @@ func downloadBlobFromRPC(client *rpc.Client, kvIndex uint64, hash common.Hash) (
...
+}
+
+func extract(text, patten string) map[string]string {
+	results := make(map[string]string)
+	re := regexp.MustCompile(patten)
+
+	matches := re.FindStringSubmatch(text)
+	if matches != nil {
+		for i, name := range re.SubexpNames() {
+			if i != 0 && name != "" { // 忽略完整匹配
```
syntrust, line 478: comments in English?

cmd/integration-test-server/main.go
```diff
@@ -175,7 +178,14 @@ func checkFinalState(state *node.NodeState) {
 			addErrorMessage("At lease one block should be mined successfully during the test.")
 		}
 		if shardState.SubmissionState.Failed > 0 {
-			addErrorMessage(fmt.Sprintf("%d submission failed during the test.", shardState.SubmissionState.Failed))
+			byDesignCount := checkByDesignFailure()
```
syntrust, line 181: I think "by design" is very misleading. They are undesired edge cases that may only occur upon intensive mining TX submissions like integration tests.

syntrust (approved):

cmd/integration-test-server/main.go
```diff
@@ -278,6 +288,202 @@ func downloadBlobFromRPC(client *rpc.Client, kvIndex uint64, hash common.Hash) (
...
+				}
+			}
+		} else if regexp.MustCompile(`Failed to submit mined result[\s\S]+invalid samples`).MatchString(logText) {
+			for kvIdx, block := range legacyKVs {
+				if block != "" && strings.Contains(logText, block) {
+					log.Warn("By design error", "block", block, "kvIdx", kvIdx, "error", "invalid samples")
+					count++
+					continue
+				}
+				delete(minedEmptyKVs, kvIdx)
```
qzhodl, line 406: we may also want to delete legacyKVs

cmd/integration-test-server/main.go
```diff
@@ -278,6 +288,202 @@ func downloadBlobFromRPC(client *rpc.Client, kvIndex uint64, hash common.Hash) (
...
+	}
+
+	if block == "" || kvIdx == "" {
+		err = errors.New(fmt.Sprintf("extract mined block and nonce fail, %s, error: wrong log format", patten))
+	}
+
+	return
+}
+
+func fetchBlocckAndDifficulty(text string) (string, string, error) {
```
qzhodl, line 442: Blocck => Block

qzhodl (approved):

Issue #332, opened 2024-11-01
Mining failed in the L2 testnet after upgrading the es-node to v0.1.16.


qzhodl:
#### System information

EthStorage node version: v0.1.16

#### Expected behaviour
Mining success

#### Actual behaviour
![image](https://github.com/user-attachments/assets/65e7e07d-18ff-44fd-bfb3-4ba8360f75f2)

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

PR #334, opened 2024-11-06
Fix `arithmetic underflow or overflow`

syntrust:
Fixes https://github.com/ethstorage/es-node/issues/332.

The issue happens when the latest version of es-node calls `miningReward` method of the storage contract before submitting the mining transaction.

The root cause is in the earlier version of the storage contract, the function `miningReward` uses the wrong `block.timestamp` and `block.number` to compute the mining timestamp in the L2 network, which may cause an underflow of unit.

The solution would be instead of aborting the transaction upon the error from `miningReward`, to use a marketable gas price, which is independent of the reward, to submit the tx anyway.

Tests been done: using a storage contract that can reproduce the error, applying the fix to the es-node can submit the tx successfully while getting a failed warning from `miningReward`.

### DIFF
--- a/ethstorage/miner/l1_mining_api.go
+++ b/ethstorage/miner/l1_mining_api.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


qzhodl (approved):
LGTM

ping-ke (approved):

PR #336, opened 2024-11-27
Fix integration test

ping-ke:
The [integration test](https://github.com/ethstorage/es-node/actions/runs/12026631901) fails with a known issue, which our pattern has not been covered. Updated our pattern to cover it.

In the previous pattern, when we mined a block with an empty blob, if the blob content was updated before "Got storage proof", we expected this mining result to fail as the blob was updated. Otherwise, we will expect the mining will succeed.
Because, after the miner "Got storage proof", we consider mining to be over and submit the mining transaction immediately. As the time between "Got storage proof" and submitting transactions is relatively short. The previous pattern believed that the possibility of data updating during this period was very small and could be ignored. 
The reason why the integration test failed this time was that data updating occurred during this time. Therefore, the pattern is modified to consider mining to be completed only after the transaction is submitted.

### DIFF
--- a/cmd/integration-test-server/main.go
+++ b/cmd/integration-test-server/main.go
@@ -350,7 +350,7 @@ func checkInvalidSamplesError() (int, error) {
 	// lvl=info msg="Get data hash"                         kvIndex=11742 hash=0x0000000000000000000000000000000000000000000000000000000000000000
 	// lvl=info msg="Downloaded and encoded"                blockNumber=4,225,672 kvIdx=11742
 	// lvl=info msg="Got storage proof"                     shard=1 block=6,906,682 kvIdx="[14613 11742]" sampleIdxsInKv="[1691 1859]"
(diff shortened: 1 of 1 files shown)


cmd/integration-test-server/main.go
```diff
@@ -399,14 +399,14 @@ func checkInvalidSamplesError() (int, error) {
...
-					log.Warn("By design error", "block", block, "kvIdx", kvIdx, "error", "invalid samples")
+					log.Warn("Known error", "block", block, "kvIdx", kvIdx, "error", "invalid samples")
 					delete(legacyKVs, kvIdx)
 					count++
 					continue
 				}
 				delete(minedEmptyKVs, kvIdx)
 			}
-		} else if strings.Contains(logText, "Mining result loop get result") {
+		} else if strings.Contains(logText, "Submit mined result done") {
```
qzhodl, line 409: Could you please provide more explanation about why the previous pattern failed to capture the issue and how the changed pattern addresses it?

qzhodl (approved):

PR #338, opened 2024-12-02
Guide to test random blob using archive service

syntrust:
Guide to test random blob using archive service, addressing milestone 3.

### DIFF
--- a/ethstorage/archiver/guide-blob-test.md
+++ b/ethstorage/archiver/guide-blob-test.md
@@ -1,0 +1,144 @@
+
+# Testing EthStorage Archive Service 
+
+## Introduction
+
+This document outlines the testing strategy and details for the blob archiver service of EthStorage. It aims to verify that the EIP-4844 blobs can be downloaded and stored by EthStorage, and also can be retrieved from EthStorage correctly while expired and pruned from the L1 Beacon client.
+
+## Environment Setup
+
+The tests will be conducted in an environment with the following services up and running:
+- An Ethereum L1 (Sepolia) RPC of an Execution Client running in archive mode
+- An Ethereum L1 (Sepolia) Beacon URL
+- An EthStorage node (es-node) with the archive service enabled
+- A proxy of Ethereum Beacon URL to mock a short blob retention period
+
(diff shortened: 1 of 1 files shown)


ethstorage/archiver/guide-blob-test.md
```diff
@@ -0,0 +1,198 @@
...
+### Running an es-node with the Archive Service Enabled
+
+Follow these steps to download the source code for es-node, build it, initialize it, and run it with the archive service enabled:
+
+```bash
+git clone https://github.com/ethstorage/es-node.git
+cd es-node 
+make
+./init-rpc.sh
+./run-rpc.sh --archiver.enabled
```
qzhodl, line 41: It might make more sense to use the proxy beacon URL in the es-node.

ethstorage/archiver/guide-blob-test.md
```diff
@@ -38,27 +52,13 @@ git clone https://github.com/ethstorage/es-node.git
 cd es-node 
 make
 ./init-rpc.sh
-./run-rpc.sh --archiver.enabled
+./run-rpc.sh  --l1.beacon http://88.99.30.186:3600 --archiver.enabled
```
qzhodl, line 55: Should this be the proxy beacon API created in the earlier step?

qzhodl (approved):

ethstorage/archiver/guide-blob-test.md
```diff
@@ -0,0 +1,198 @@
...
+
+To download and run a proxy for the Beacon client, execute the following commands:
+
+```bash
+git clone https://github.com/ethstorage/beacon-api-wrapper.git
+cd beacon-api-wrapper
+go run cmd/main.go -r 1800
+```
+
+This proxy functions like a standard Beacon API, except that it has a much shorter blob retention period—30 minutes in this case. Consequently, when a `blob_sidecars` request is made for blobs older than 150 slots, it will return an empty list: `{"data":[]}`.
```
blockchaindevsh, line 42: It's better to make it explicit that 30 min correspond to the `1800` parameter, which is retention period in seconds.

blockchaindevsh (approved):

PR #340, opened 2024-12-05
Archiver test doc update

syntrust:
Fix comments during the demo. Major changes include:

- Reorg the table of contents
- Move the es-node setup to a later point to compare results with and without the archive service.
- Correct the incorrect command
- Update the beacon-api-wrapper to use epoch as the unit of retention time (per EIP-4844, which defines the retention period as 4096 epochs).
- Address inconsistencies in titles, etc.

Changes verified in the test environment.

### DIFF
--- a/ethstorage/archiver/guide-devnet.md
+++ b/ethstorage/archiver/guide-devnet.md
@@ -9,9 +9,10 @@
    - [Fill Out Environment Variables](#fill-out-environment-variables)
 2. [L1 Setup](#l1-setup)
    - [Starting L1](#starting-l1)
+   - [Running a Proxy of L1 Beacon to Mock Short Retention Period of Blobs](#running-a-proxy-of-l1-beacon-to-mock-short-retention-period-of-blobs)
+3. [Deploying Contracts](#deploying-contracts)
    - [Deploying EthStorage Contracts](#deploying-ethstorage-contracts)
-   - [Deploy BatchInbox](#deploy-batchinbox)
-   - [Running a Proxy of L1 Beacon](#running-a-proxy-of-l1-beacon)
+   - [Deploying BatchInbox Contract](#deploying-batchinbox-contract)
 3. [Running EthStorage Node](#running-ethstorage-node)
    - [Installation](#installation)
    - [Initialization](#initialization)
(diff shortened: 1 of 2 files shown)


ethstorage/archiver/guide-devnet.md
```diff
@@ -342,17 +291,81 @@ make op-node
...
+```bash
+curl -s http://localhost:5052/eth/v1/beacon/genesis | jq -r '.data.genesis_time'
+
+1732529739
+
+export GENESIS_TIME=1732529739 // replace with the actual timestamp
+```
+Note: Before proceeding to the next step of launching the es-node, ensure that at least 2 epochs (approximately 13 minutes) have passed since the EthStorage contracts were deployed in [this step](#deploying-ethstorage-contracts), as the es-node needs to read the finalized states of the contract.
+
+Finally, run the es-node:
```
qzhodl, line 338: Should we mention that at least 2 epochs need to pass after deploying the storage contract before launching the es-node? Qi encountered this issue while following the documentation.

ethstorage/archiver/guide-devnet.md
```diff
@@ -342,17 +291,81 @@ make op-node
...
 1. P2P is disabled so that it can only sync data from L1.
 2. The L1 beacon URL is directed to the beacon proxy, where blobs expire quickly.
-3. The beacon archiver is configured to point to the es-node archive service. 
 
-## Verifying Sync Status
+You can observe that the validator node is unable to sync properly with "derivation failed" error due to "failed to fetch blobs". 
+
+Now, navigate to the parent directory in preparation for the next steps.
+
+## Running EthStorage Node
```
qzhodl, line 305: If I understand correctly, this is the only instance of es-node in the network, and the downloader only attempts to retrieve blobs from the latest finalized block, while the P2P module syncs older blobs from other peers. Since this es-node was launched after the L2 node, there’s a chance that the es-node might fail to sync blobs. I would recommend launching the es-node before starting the L2 node.

syntrust: Yes that makes sense. Changed the order again.

qzhodl (approved):

blockchaindevsh (approved):

Issue #344, opened 2024-12-06
Remove --l1.beacon-based-time and --l1.beacon-based-slot flags


qzhodl:
# Rationale

We can use /eth/v1/beacon/genesis to query timestamp of slot 0.

PR #345, opened 2024-12-09
Remove beacon base options

syntrust:
Fix https://github.com/ethstorage/es-node/issues/344

Tests have been done in a devnet environment where es-node can download blobs correctly and serves as DA for an op node to derive.


qzhodl (commented):
LGTM

qzhodl (approved):

PR #355, opened 2025-01-02
Add Chainid to p2p filter

ping-ke:
**Issue:** 
Currently, we use l1.chain_id to filter ENR in the discovery UDPv5 table, which cannot filter out the es-node ENR that has been connected before. Also, the run*.sh does not set the l1.chain_id, and the es-node will use 1 as the default value. 
So now, we only use the contract address as a filter in the SyncClient.AddPeer filters out Es-nodes that do not belong to the same network. However, when we change SWC from Alpha to Beta, the contract address will not change. So, the Alpha and Beta ES-node networks will join together if any Beta ES-node does not clear its discovery DB (`esnode_discovery_db`) before it starts. 

**How to fix:**
1. Use L2 chain ID to generate ENRs instead of L1 Chain ID
2. Filter all ENRs that do not have the same L2 Chain ID, so networks that do not use the same L2 Chain ID will not join together through discovery

Note: 
1. After this change, the Bootnode ENR will be changed, 
2. As `esnode_peerstore_db` keeps the addresses and shards info for es-nodes they connected before, so es-node without the latest code can re-connect to Bootnode or es-nodes updated to the latest code. (Sepolia test)

**How to test:** 
1. New Bootnode **`with`** this chain ID change + new es-node **`with`** this chain ID change should connect and sync.
2. New Bootnode **`with`** this chain ID change + new es-node **`without`** this chain ID change should not connect and sync.
3. Old Bootnode **`without`** this chain ID change + new es-node **`with`** this chain ID change should not connect and sync.
4. Old Bootnode **`switch`** to this chain ID change **`with`** or **`without`** clear `esnode_discovery_db` and `esnode_peerstore_db`, old es-node **`without`** this chain ID change still can connect and sync.
5. Old es-node **`switch`** to this chain ID change **`without`** clear `esnode_discovery_db` and `esnode_peerstore_db`, it should connect to old Bootnode **` with`** or **` without`** this chain ID change.

Note:
New bootnode and new es-node means this bootnode/es-node using an IP+Port that has not been used before.
Old bootnode and es-node means using this bootnode/es-node using an IP+Port that has been used before, and other es-nodes have recorded it (IP+Port) in their esnode_peerstore_db.

**How to update nodes:**
1. Bootnode update to the latest code, and start Bootnode again.
2. Copy Bootnode ENR and update run-L2.sh and run.sh
3. Other Nodes pull the latest code and restart again

Note: For nodes which switch to chain ID change without clear `esnode_peerstore_db` can still directly connect to the es-nodes which they have connected before, not matter those nodes have swtich chain ID change or not.


ethstorage/p2p/discovery.go
```diff
@@ -365,6 +365,30 @@ func (n *NodeP2P) DiscoveryProcess(ctx context.Context, log log.Logger, l1ChainI
 	}()
 
 	pstore := n.Host().Peerstore()
+
+	// clear pstore from DB if the version is not match
+	// the DB path is set by p2p.peerstore.path flag, and the default value is esnode_peerstore_db
+	go func() {
```
qzhodl, line 371: What is the purpose of this function, and in what scenarios would we need to increase the p2pversion?

qzhodl:
Is there a note that outlines what boot node and full node operators should do for the beta testnet if they take this PR? Additionally, should we update our runL2.sh script to include the new ENR?

ethstorage/p2p/discovery.go
ping-ke: Update function description

qzhodl (approved):
LGTM

ethstorage/p2p/discovery.go
```diff
@@ -208,8 +208,8 @@ func FilterEnodes(log log.Logger, l1ChainID uint64) func(node *enode.Node) bool
 			return false
 		}
 		// check chain ID matches
-		if l1ChainID != dat.ChainID {
-			log.Trace("Discovered node record has no matching chain ID", "node", node.ID(), "got", dat.ChainID, "expected", l1ChainID)
+		if l1ChainID != dat.L1ChainID {
```
syntrust, line 211: Just wondering, could we use L2ChainID here to serve the same purpose?

syntrust:
Sounds like a hardfork to L2 testnet. Any impact on the L1 testnet? Does anything need to change to the storage providers' guide? Specifically, based on https://github.com/ethstorage/ethstorage-doc/pull/37?

> Note: After this change, the Bootnode ENR will be changed, so all the Es-nodes in the network need to sync to the latest code and restart.
> 
> How to test:
> 
> 1. Bootnode **`with`** this chain ID change + es-node **`with`** this chain ID change should connect and sync.
> 2. Bootnode **`with`** this chain ID change + es-node **`without`** this chain ID change should not connect and sync.
> 3. Bootnode **`without`** this chain ID change + es-node **`with`** this chain ID change should not connect and sync.
> 4. Bootnode **`switch`** to this chain ID change without clear esnode_discovery_db and esnode_peerstore_db, es-node **`without`** this chain ID change should not connect and sync.
> 5. es-node **`switch`** to this chain ID change without clear esnode_discovery_db and esnode_peerstore_db, it should
>    connect to Bootnode **` with`** this chain ID change and it should not connect to Bootnode **` without`** this chain ID change.
> 
> How to update nodes:
> 
> 1. Bootnode update to the latest code, remove `esnode_discovery_db` and `esnode_peerstore_db` folders, and start Bootnode again.
> 2. Copy Bootnode ENR and update run-L2.sh
> 3. Other Nodes pull the latest code and restart again
> 
> Note: For nodes that have completed synchronization, there is no impact on them. For nodes that are synchronizing, they can only sync **` with`** nodes with the same p2pVersion.

syntrust:
Due to potential backward incompatibility issues on the current testnets, especially the L1 public testnet, we may need to defer merging this PR into the upcoming release.

syntrust (approved):
Note that the ENR in run*.sh should be updated to the latest after bootnode deployment.

qzhodl (approved):

Issue #356, opened 2025-01-03
Mining gas cost estimation needs to consider l1 cost


syntrust:
Regarding estimate tx fee (Execution gas fee + L1 data fee), the [OP's document](https://docs.optimism.io/builders/app-developers/transactions/estimates#l1-data-fee) explains how to do it.

PR #360, opened 2025-01-17
Add L1 data fee when estimating mining transaction profit

syntrust:
Fixes https://github.com/ethstorage/es-node/issues/356. 

Also, implement the checking of the gas fee against the tx fee cap (e.g., `--RPC.txfeecap` as in geth) before submitting the tx to avoid a tx fail on the chain.

Tests:

Deployed a contract to SWC beta and did the mining tests with:
```bash
env ES_NODE_SIGNER_PRIVATE_KEY=<ES_NODE_SIGNER_PRIVATE_KEY> ./integration_tests/run_tests.sh
``` 

It also successfully mined as a member node in the SWC beta network.


ethstorage/miner/l1_mining_api.go
```diff
@@ -112,12 +115,59 @@ func (m *l1MiningAPI) SubmitMinedResult(ctx context.Context, contract common.Add
...
+
 	reward, err := m.GetMiningReward(rst.startShardId, rst.blockNumber.Int64())
 	if err != nil {
 		m.lg.Warn("Query mining reward failed", "error", err.Error())
 	}
 	if reward != nil {
-		profitableGasFeeCap := new(big.Int).Div(new(big.Int).Sub(reward, cfg.MinimumProfit), new(big.Int).SetUint64(estimatedGas))
+		m.lg.Info("Query mining reward done", "reward", reward)
+		costCap := new(big.Int).Sub(reward, cfg.MinimumProfit)
+		costCap = new(big.Int).Add(costCap, extraCost)
```
qzhodl, line 169: should be `costCap - extraCost`?

ethstorage/miner/l1_mining_api.go
```diff
@@ -112,12 +115,59 @@ func (m *l1MiningAPI) SubmitMinedResult(ctx context.Context, contract common.Add
...
+		To:        &contract,
+		Value:     common.Big0,
+		Data:      calldata,
+	}
+
+	unsignedTx := types.NewTx(rawTx)
+	extraCost := new(big.Int)
+	// Add L1 data fee as tx cost when es-node is deployed as an L3
+	if m.rc != nil {
+		l1fee, err := m.getL1Fee(ctx, unsignedTx)
```
qzhodl, line 153: Is there any difference between unsignedTx and signedTx? If so, I believe using signedTx would be more accurate

integration_tests/run_tests.sh
```diff
@@ -26,15 +26,15 @@ if [ -z "$ES_NODE_STORAGE_L1CONTRACT_CLEF" ]; then
 fi
 # A newly deployed contract is required for each run for miner test, with zkp verifier of mode 2
 if [ -z "$ES_NODE_STORAGE_L1CONTRACT" ]; then
-  export ES_NODE_STORAGE_L1CONTRACT=0xe8F0898cbA701E677970DB33404A817Ff42D4499
+  export ES_NODE_STORAGE_L1CONTRACT=0x517ad0ba959f3556930c9Bc483B454584F7e11df
```
qzhodl, line 29: Does this test adequately cover the changes introduced in this PR?

ethstorage/miner/l1_mining_api.go
syntrust: Using signed tx may return a different fee from unsigned, but it looks like the function requires unsigned tx: https://github.com/ethereum-optimism/optimism/blob/39e9f1a4693912af8bdef428d246c15a6cf44ec7/packages/contracts-bedrock/src/L2/GasPriceOracle.sol#L55

ethstorage/miner/l1_mining_api.go
syntrust: fixed.

integration_tests/run_tests.sh
syntrust: add unit tests

qzhodl (approved):
LGTM

ping-ke (approved):

PR #373, opened 2025-03-06
upgrade geth to Pectra(v1.15.5)

syntrust:
The main reason to do this is the block header structure change causes "RandaoLib: header hash mismatch"

This is the new field added to Pectra:
```
// RequestsHash was added by EIP-7685 and is ignored in legacy headers.
	RequestsHash *common.Hash `json:"requestsHash" rlp:"optional"`
```


ping-ke (approved):

ethstorage/eth/beacon_client.go
```diff
@@ -11,8 +11,8 @@ import (
 	"net/url"
 	"strconv"
 
-	"github.com/crate-crypto/go-proto-danksharding-crypto/eth"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/protolambda/go-kzg/eth"
```
qzhodl, line 15: can we use kzg4844.CalcBlobHashV1?

ethstorage/log/cli.go
```diff
@@ -53,22 +54,40 @@ func (cfg CLIConfig) Check() error {
...
-	// Set the root handle to what we have configured. Some components like go-ethereum's RPC
-	// server use log.Root() instead of being able to pass in a log.
-	log.Root().SetHandler(handler)
-	logger := log.New()
-	logger.SetHandler(handler)
+	l, err := LevelFromString(cfg.Level)
+	if err != nil {
+		panic(err)
+	}
+	h := log.NewTerminalHandlerWithLevel(os.Stdout, l, cfg.Color)
```
qzhodl, line 89: How can we support other format like "logfmt", "json", etc.

ethstorage/eth/beacon_client.go
syntrust: done

ethstorage/log/cli.go
syntrust: Filed https://github.com/ethstorage/es-node/issues/374

qzhodl (approved):

PR #381, opened 2025-03-12
es-node tool to fix a specific kv entry

syntrust:
An es-node subcommand to download a blob from an RPC node and update it in local storage. This tool can fix a data entry that was accidentally incorrect.

Usage example:
```bash
env ES_NODE_STORAGE_MINER=0x4b07e60490BE95c03d41F63203De9b51657A4C1a ./sync.sh --kv_index=2012299
```

The hash can be checked and verified using:
```bash
cd cmd/es-utils
go build
./es-utils meta_read --filename ../../es-data/shard-0.dat  --kv_idx=2012299  --kv_entries=8192 --kv_size 131072 --chunk_size 131072
```

incorrect  hash
>01abcdf2f4528e806a2a89c452e1c38209180dfd4db4a8c08000000000000000

correct hash
>01aa945c93ead8b4630b48e3b380864b0c663c2383a4d6d48000000000000000

### DIFF
--- a/cmd/es-node/main.go
+++ b/cmd/es-node/main.go
@@ -18,7 +18,9 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/ethstorage/go-ethstorage/ethstorage"
+	"github.com/ethstorage/go-ethstorage/ethstorage/eth"
 	"github.com/ethstorage/go-ethstorage/ethstorage/flags"
 	eslog "github.com/ethstorage/go-ethstorage/ethstorage/log"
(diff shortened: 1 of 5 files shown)


qzhodl (approved):
LGTM

qzhodl:
Can we specify an kvIndex range to sync?

syntrust:
I think a quick implementation and simple to use is preferable, especially in the current case. We can add array support later if necessary.

> Can we specify a kvIndex range to sync?

ping-ke:
Is there any way to regular check we have save the correct data?
Is it better to do this using es-utils?

syntrust:
> Is there any way to regular check we have save the correct data? Is it better to do this using es-utils?

es-utils is more like an internal toolkit, plus we need the same group of flags used in es-node.

As for a regular way to check data, an [issue](https://github.com/ethstorage/es-node/issues/384) is filed for further discussion.

ping-ke (approved):

PR #392, opened 2025-04-08
Data check 

syntrust:
This PR implements feature https://github.com/ethstorage/es-node/issues/384.

## Usage
Refer to https://github.com/ethstorage/es-node/tree/data-check/ethstorage/scanner#usage

## Tests

Test results on L1 with the default settings (check meta) shows 8192 kvs can be checked within about 350ms:
`--scanner.mode 1 --scanner.interval 3 --scanner.batch-size 8192`

```
INFO [04-08|10:51:51.077] Scanner update loop received event - all shards done.
INFO [04-08|10:51:51.077] Scanner started                          interval=3m0s
INFO [04-08|10:51:51.077] Scan batch started                       mode=1
INFO [04-08|10:51:51.103] Scanner: query KV entry count done       kvEntryCount=2,012,738
INFO [04-08|10:51:51.137] Scanner: query local KV entries done     localKVs="[0 1 2... 2012735 2012736 2012737]"
INFO [04-08|10:51:51.423] Scanner: query KV meta done              kvsInBatch="[0 1 2... 8189 8190 8191]"
INFO [04-08|10:51:51.428] Scanner: scan batch done                 from=0 to=8191 count=8192 nextKvIndex=8192
INFO [04-08|10:51:51.428] Scan batch done                          duration=351.132629ms
INFO [04-08|10:54:51.078] Scan batch started                       mode=1
INFO [04-08|10:54:51.104] Scanner: query KV entry count done       kvEntryCount=2,012,738
INFO [04-08|10:54:51.139] Scanner: query local KV entries done     localKVs="[0 1 2... 2012735 2012736 2012737]"
INFO [04-08|10:54:51.426] Scanner: query KV meta done              kvsInBatch="[8192 8193 8194... 16381 16382 16383]"
INFO [04-08|10:54:51.431] Scanner: scan batch done                 from=8192 to=16383 count=8192 nextKvIndex=16384
INFO [04-08|10:54:51.431] Scan batch done                          duration=352.769547ms
INFO [04-08|10:57:51.078] Scan batch started                       mode=1
INFO [04-08|10:57:51.103] Scanner: query KV entry count done       kvEntryCount=2,012,738
INFO [04-08|10:57:51.114] Scanner: query local KV entries done     localKVs="[0 1 2... 2012735 2012736 2012737]"
INFO [04-08|10:57:51.392] Scanner: query KV meta done              kvsInBatch="[16384 16385 16386... 24573 24574 24575]"
[...]

PR #395, opened 2025-04-23
 Add FetchBlob to SyncClient to support fetching a single blob from the network via p2p

ping-ke:
Add FetchBlob to SyncClient to support fetching a single blob from the network via p2p, it will use by [pr 392](https://github.com/ethstorage/es-node/pull/392/files).

interface: 
func FetchBlob(index uint64, commit common.Hash) ([]byte, error)

### DIFF
--- a/ethstorage/p2p/protocol/syncclient.go
+++ b/ethstorage/p2p/protocol/syncclient.go
@@ -625,7 +625,7 @@ func (s *SyncClient) FetchBlob(kvIndex uint64, commit common.Hash) ([]byte, erro
 			if val.BlobIndex != kvIndex {
 				continue
 			}
-			if bytes.Equal(val.BlobCommit[0:ethstorage.HashSizeInContract], commit[0:ethstorage.HashSizeInContract]) {
+			if !bytes.Equal(val.BlobCommit[0:ethstorage.HashSizeInContract], commit[0:ethstorage.HashSizeInContract]) {
 				log.Warn("FetchBlob failed", "peer", pr.ID(), "expected commit", commit.Hex(), "actual commit", val.BlobCommit.Hex())
 				continue
 			}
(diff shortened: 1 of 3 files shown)


ethstorage/p2p/protocol/syncclient.go
```diff
@@ -596,15 +596,59 @@ func (s *SyncClient) RequestL2Range(start, end uint64) (uint64, error) {
...
 			return 0, err
 		}
-		return id, nil
+
+		return uint64(len(inserted)), nil
 	}
 	return 0, fmt.Errorf("no peer can be used to send requests")
 }
 
+func (s *SyncClient) FetchBlob(kvIndex uint64, commit common.Hash) ([]byte, error) {
```
syntrust, line 609: It would be nice to return nil instead of []byte{}

ethstorage/p2p/protocol/syncclient.go
```diff
@@ -596,15 +596,59 @@ func (s *SyncClient) RequestL2Range(start, end uint64) (uint64, error) {
...
+	if len(s.peers) == 0 {
+		return []byte{}, fmt.Errorf("no peer can be used to send requests")
+	}
+	for _, pr := range s.peers {
+		var packet BlobsByListPacket
+		var payload *BlobPayload = nil
+
+		_, err := pr.RequestBlobsByList(rand.Uint64(), s.storageManager.ContractAddress(), kvIndex/s.storageManager.KvEntries(), []uint64{kvIndex}, &packet)
+		if err != nil {
+			log.Warn("FetchBlob failed", "error", err)
```
syntrust, line 619: Should `continue` after warning?

ethstorage/p2p/protocol/syncclient.go
```diff
@@ -596,15 +596,59 @@ func (s *SyncClient) RequestL2Range(start, end uint64) (uint64, error) {
...
+		}
+
+		for _, val := range packet.Blobs {
+			if val.BlobIndex != kvIndex {
+				continue
+			}
+			if val.BlobCommit.Cmp(commit) != 0 {
```
syntrust, line 626: We should compare the part of `commit[0:HashSizeInContract]` instead of all hash

syntrust (approved):

qzhodl (approved):

PR #407, opened 2025-06-27
check miner role

syntrust:
Check if the miner is in the miner's role (whitelisted) when es-node is started.

Tests have been done when starting es-node:

Based on storage contract version:

**v0.1.1**
ignore role check

**v0.1.2**
Happy paths:
-  Miner role enforced / Miner role granted
-  Miner role not enforced

Failed to start es-node:
-  Miner role enforced / Miner role not granted

### DIFF
--- a/ethstorage/miner/l1_mining_api.go
+++ b/ethstorage/miner/l1_mining_api.go
@@ -19,10 +19,12 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/ethstorage/go-ethstorage/ethstorage"
 	"github.com/ethstorage/go-ethstorage/ethstorage/eth"
+	"golang.org/x/mod/semver"
 )
 
 const (
-	gasBufferRatio = 1.2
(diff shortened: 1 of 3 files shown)


ethstorage/miner/l1_mining_api.go
```diff
@@ -42,6 +42,34 @@ type l1MiningAPI struct {
 	lg log.Logger
 }
 
+// Pre-check if the miner has been whitelisted before actually mining
+func (m *l1MiningAPI) CheckMinerRole(ctx context.Context, contract, miner common.Address) error {
+	enforced, err := m.PollingClient.ReadContractField("enforceMinerRole", nil)
+	if err != nil {
+		return fmt.Errorf("failed to query enforceMinerRole(): %w", err)
```
qzhodl, line 49: The new version of the ES Node should remain compatible with the existing storage contract.

qzhodl (approved):

PR #437, opened 2025-09-04
Dashboard: Load mined event from DB instead of refetch again from chain.

ping-ke:


cmd/dashboard/config.json
```diff
@@ -1,14 +1,14 @@
 [
   {
-    "name":"Sepolia",
+    "name":"EthStorage",
     "type": "l1",
     "rpc":"http://65.108.230.142:8545",
-    "chainID": 3333,
-    "contract":"0x804C520d3c084C805E37A35E90057Ac32831F96f",
-    "startNumber": 5528000
+    "chainID": 3335,
```
qzhodl, line 6: should be 3333?

PR #443, opened 2025-09-08
Archive tests using new public testnet

syntrust:
1.  archiver-guide-blob-test.md updated based on latest test
2. new doc archiver-guide-sepolia-sync-gamma.md describe the process of test done with gamma testnet
3. fix "invalid indices": optionally check blob index
4. minor changes 

Tests done with a gamma node syncing.

### DIFF
--- a/ethstorage/scanner/scanner.go
+++ b/ethstorage/scanner/scanner.go
@@ -24,8 +24,6 @@ type Scanner struct {
 	running   bool
 	mu        sync.Mutex
 	lg        log.Logger
-	errorCh   chan scanError
-	statsCh   chan stats
 	scanStats ScanStats
 }
 
@@ -46,8 +44,6 @@ func New(
 		ctx:       cctx,
 		cancel:    cancel,
 		lg:        lg,
-		errorCh:   make(chan scanError, 10),
-		statsCh:   make(chan stats, 10),
 		scanStats: ScanStats{0, 0},
 	}
 	scanner.wg.Add(1)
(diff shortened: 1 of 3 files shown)


docs/archiver/archiver-guide-sepolia-sync-gamma.md
```diff
@@ -0,0 +1,233 @@
...
+INFO [09-16|09:26:24.367] Blob archiver API request handled        took(s)=0.456
+INFO [09-16|10:27:09.649] Blob archiver API request                url="/eth/v1/beacon/blob_sidecars/8518431?indices=1"
+INFO [09-16|10:27:10.063] Blob archiver API request handled        took(s)=0.414
+INFO [09-16|12:31:19.786] Blob archiver API request                url="/eth/v1/beacon/blob_sidecars/8522289?indices=5"
+INFO [09-16|12:31:20.362] Blob archiver API request handled        took(s)=0.576
+```
+
+From the logs, we can see that the op-node keeps converting batches into payload attributes, calling the op-geth to work on the payload to convert it into blocks, and inserting the blocks as the chain head.
+
+Additionally, you may notice the age of derived blocks is over 12 hours, and blob archiver API requests were handled by the es-node, meaning the blobs were retrieved from EthStorage since they were pruned by the L1 Beacon client (mocked).
```
qzhodl, line 218: It’s still unclear under what conditions we can reproduce this and make op-node fetch the blob from the archive API instead of the mock beacon API.

PR #453, opened 2025-09-16
Fix failed to estimate gas: max fee per gas less than block base fee

syntrust:
Fix issue from mainnet:

```log
INFO [09-15|03:02:02.260] Query baseFee done                       baseFee=157,624,066
INFO [09-15|03:02:02.321] Query gas tip cap done                   gasTipGap=10
INFO [09-15|03:02:02.321] Suggested gas fee cap                    gasFeeCap=157,624,076
ERROR[09-15|03:02:02.514] Estimate gas failed                      error="failed with 44912112 gas: max fee per gas less than block base fee: address 0xA8E33F00B63a432C30C732da6670DE5eDCC16980, maxFeePerGas: 157624076, baseFee: 160487922: no revert data"
```
The issue may caused by using the latest block's base fee + tip to estimate gas limit, while the on-chain base fee fluctuates between blocks. A safe way would be `base fee * 2 + tip`.

Tests been done with successful mining and drop L2 locally, and L1 testnet.


ethstorage/miner/l1_mining_api.go
```diff
@@ -350,24 +354,23 @@ func checkGasPrice(
...
-				fmtEth(reward), fmtEth(new(big.Int).Mul(new(big.Int).SetUint64(estimatedGas), gasFeeCap)), fmtEth(extraCost), fmtEth(profit), fmtEth(minProfit))
+				fmtEth(reward), fmtEth(new(big.Int).Mul(new(big.Int).SetUint64(estimatedGas), gasPrice)), fmtEth(extraCost), fmtEth(profit), fmtEth(minProfit))
 			return nil, errDropped{reason: droppedMsg}
 		}
+		// Cap the gas fee to be profitable
 		if !useConfig {
 			gasFeeCapChecked = profitableGasFeeCap
 			lg.Info("Using profitable gas fee cap", "gasFeeCap", gasFeeCapChecked)
 		}
-	} else {
```
qzhodl, line 365: Why do we delete the `else` case?

syntrust: Because gasFeeCap is already baseFee*2 + tip

iteyelmp (approved):

PR #468, opened 2025-10-09
refactor: use reflect.TypeFor

yajianggroup:
Try to use better api reflect.TypeFor instead of reflect.TypeOf when we have known the type.

More info https://github.com/golang/go/issues/60088


yajianggroup:
@qzhodl  @syntrust  Thanks for your review!

If there's anything need I to do, please feel free to let me know.

PR #478, opened 2025-11-24
update archive docs

syntrust:
Based on actual test done with the updated API: `/eth/v1/beacon/blobs/$SLOT?versioned_hash=$VERSIONED_HASH`

### DIFF
--- a/docs/archiver/archiver-guide-blob-test.md
+++ b/docs/archiver/archiver-guide-blob-test.md
@@ -121,13 +121,10 @@ With block explorer, you can find the versioned hash of the blob you just upload
 ```
 "https://sepolia.etherscan.io/tx/$TX_HASH#blobs"
 ```
-Export value for later use:
-```
+Export the value for later use:
+```bash
 export VERSIONED_HASH=0x0158420bd7b1f3c04a097694235c52659f31b479018ac56b2545727fab33712d
 ```
-
-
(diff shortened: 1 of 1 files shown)


docs/archiver/archiver-guide-mainnet-sync.md
```diff
@@ -0,0 +1,205 @@
...
+4. **[Conclusion](#conclusion)**
+
+
+## Introduction
+
+This guide walks you through spinning up a QuarkChain L2 node on mainnet using the EthStorage archive service.
+
+QuarkChain L2, built on OP Stack, publishes batches as blobs to its [BatchInbox contract](https://etherscan.io/address/0xf62e8574B92dc8764c5Ad957b5B0311595f5A3f9) on Ethereum mainnet. EthStorage stores those blobs as a long-term data-availability layer.
+
+The objective is to verify that the L2 node can derive blocks correctly from blobs fetched via EthStorage.
```
blockchaindevsh: Maybe we should be more clear whether the purpose here is:
1. walks you through spinning up a QuarkChain L2 node on mainnet using the EthStorage archive service; or
2. The objective is to verify that the L2 node can derive blocks correctly from blobs fetched via EthStorage.

as they sound like two different purpose to me.

(I guess 2 is the real purpose?)

PR #506, opened 2026-03-11
Fix `FailedToObtainBlockhash` in case of empty slots 

syntrust:
### Issue
The `FailedToObtainBlockhash` issue was reproduced:
```
INFO [03-09|14:19:15.072] Calculated a valid hash                  shard=0 block=10,414,872 timestamp=1,773,062,352 randao=8651fa..436b57 nonce=170,011 hash0=d04672..40e44a hash1=000000..8ce8a8 sampleIdxs="[2285954122 5178046772]"
...
INFO [03-09|14:19:40.388] Mining result loop get result            shard=0 block=10,414,872 nonce=170,011
...
ERROR[03-09|14:19:40.399] Estimate gas failed                      error="execution reverted: ErrorData: 0x4882d76b: StorageContract_FailedToObtainBlockhash()"
```

### Cause

The chain head stayed as 10,414,872 for 3 slots during the block mined and tx estimated. The fix in https://github.com/ethstorage/es-node/pull/487 only waits for 1 slot at most.

We need the block number updates to avoid empty blockhash.

### Solution

This PR replaces slot-based waiting with a channel-based synchronization that ensures mining results are only submitted after the L1 head has advanced past the mined block. 

### Tests
Tests done in local devnet:

Waiting for L1 head to advance in the case of an empty slot:
```
INFO [03-11|11:59:26.576] Calculated a valid hash                  shard=37 block=10,424,641 timestamp=1,773,201,564 randao=3756ab..928ca0 nonce=541,929 hash0=ea0224..c03870 hash1=000001..326081 sampleIdxs="[1226864 1215777]"
...
INFO [03-11|11:59:40.490] Mining result loop get result            shard=37 block=10,424,641 nonce=541,929
INFO [03-11|11:59:40.490] L1 head not advanced since mined block, keep waiting mined=10,424,641 latest=10,424,641
INFO [03-11|11:59:40.490] Waiting for L1 head to advance           mined=10,424,641
INFO [03-11|11:59:49.055] headUpdateCh updated                     new=10,424,642
INFO [03-11|11:59:49.055] L1 head advanced since mined block       mined=10,424,641 latest=10,424,642
INFO [03-11|11:59:49.740] Composed calldata   ...                    
```

Normal case:
```
[...]

PR #507, opened 2026-03-18
Estimate gas at pending block to fix `FailedToObtainBlockhash`

syntrust:
This PR is another solution as an alternative to https://github.com/ethstorage/es-node/pull/506.

## Rationale

The EVM specification requires that `BLOCKHASH` can only access the most recent 256 generated blocks, i.e., `[block.number - 256, block.number - 1]`, excluding the current block itself.

`eth_estimateGas` executes in the context of the latest chain head block `N` [by default](https://github.com/ethereum/go-ethereum/blob/ab357151da881f0e7dffd30b639c9dcdcdd4cc10/internal/ethapi/api.go#L923), so `block.number = N`.
If `BLOCKHASH(N)` is called, since `N >= block.number`, it returns `0`.

When explicitly using the `pending` block tag, the execution context changes. In Geth, `block.number = N + 1`, i.e., the next block being constructed. Therefore, `BLOCKHASH(N)` becomes valid (since `N = block.number - 1`).

In real execution, the transaction is included in block `N + k` (where `k ≥ 1`), so `block.number = N + k`. In the common case where `k = 1`, this matches the `pending` context.

## Tests

Tests have been performed on the local testnet using extra temporary debug messages to confirm that the mining tx is completed successfully in the case of mining and estimation occurring in the same slot.

### DIFF
--- a/ethstorage/miner/l1_mining_api.go
+++ b/ethstorage/miner/l1_mining_api.go
@@ -161,16 +161,6 @@ func (m *l1MiningAPI) SubmitMinedResult(ctx context.Context, contract common.Add
 		return common.Hash{}, fmt.Errorf("failed to estimate gas: %v", errMessage)
 	}
 	m.lg.Info("Estimated gas done", "gas", estimatedGas)
-	// query current block number for debug
-	currentBlock, err := m.BlockNumber(ctx)
-	if err != nil {
-		m.lg.Warn("Failed to query current block number", "error", err)
-	} else {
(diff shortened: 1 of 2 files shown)


qzhodl:
Do we also need to remove the logic that waits for one slot before calling estimateGas?

syntrust:
> Do we also need to remove the logic that waits for one slot before calling estimateGas?

Done.

qzhodl (approved):

iteyelmp (approved):

PR #518, opened 2026-08-29
chore: fix shard manager helper name

zloglevel:
Rename the package-private `findShardManaager` helper to `findShardManager` and update its two callers.

This corrects the helper name without changing shard lookup behavior or any exported API.

Tested with:

```text
go test ./ethstorage
```


zloglevel:
@qzhodl Could you please take a look when you have a chance? Thanks!

